PR #15, opened 2026-02-05
Fix double scrollbar issue in documentation site

WuMingDao:
Problem:
The documentation site was displaying two vertical scrollbars on the right side, creating a confusing user experience where users could scroll using either scrollbar.

fix before:
image

fix after:
image

Root Cause:
There were two nested scrollable containers:
1. The body element had `overflow-y: scroll` (global.css:26)
2. The #main-content container had `overflow-y-auto` (MainLayout.astro:29)

This created a double-scrolling scenario where both the body and the main content area could scroll independently.

Solution:
Changed body overflow from `overflow-y: scroll` to `overflow: hidden`, making #main-content the single source of scrolling for the entire application.

Why This Approach:
This follows the modern app-layout pattern where:
- Fixed sidebar (SidebarNav) remains visible at all times
- Fixed theme toggle stays in the top-right corner
- Only the main content area scrolls
- Consistent with modern web applications (VS Code, GitHub, etc.)

This approach provides better UX by having a single, predictable scroll container while maintaining the fixed positioning of navigation elements.

### DIFF
--- a/packages/kumo-docs-astro/src/styles/global.css
+++ b/packages/kumo-docs-astro/src/styles/global.css
@@ -21,7 +21,7 @@ html {
   font-display: optional;
 }
 
-/* Prevent layout shift on page load */
+/* Prevent layout shift on page load - scrolling handled by #main-content container */
 body {
-  overflow-y: scroll;
+  overflow: hidden;
 }


geoquant:
think this one resolved your fix, thank you:
https://github.com/cloudflare/kumo/pull/18

PR #17, opened 2026-02-05
fix(Table): Make checkbox selection interactive and fix click handling

mattrothenberg:
## Summary

- Fix Table checkbox demos on docs site - checkboxes now work interactively
- Add `indeterminate` prop support to `Table.CheckHead` and `Table.CheckCell`
- Add default `w-10` (40px) width to checkbox cells
- Expand checkbox hit area using CSS pseudo-element
- Add dropdown menu to full example's action button

https://github.com/user-attachments/assets/edc80e25-683c-49e2-8970-a49787f03d70

## Problem

The Table component demos at https://kumo-ui.com/components/table/ had non-functional checkboxes:
1. "Selected Row" and "Full Example" tables didn't show row highlights when selected
2. Checkboxes didn't work at all - they were static with no state management

### Component changes (`table.tsx`)
- Added `indeterminate` prop to `Table.CheckHead` and `Table.CheckCell` for partial selection state
- Added `w-10` default width to checkbox cells for consistent sizing
- Used CSS pseudo-element (`before:-inset-3`) to expand the checkbox hit area without needing wrapper hacks

### Demo changes (`TableDemo.tsx`)
- Added proper React state management (`useState<Set<string>>`) to all checkbox demos
- Wired up `onValueChange` handlers for interactive selection
- Added `indeterminate` state to header checkboxes
- Removed redundant colgroup width for checkbox column (component handles it now)
- Added DropdownMenu to the action button with View, Edit, Delete options

## Open Question

**Is the `variant="selected"` API on `Table.Row` weird?** Would a boolean make more sense here?

```tsx
<Table.Row variant={selectedIds.has(row.id) ? "selected" : "default"}>
  <Table.CheckCell checked={selectedIds.has(row.id)} />
```

Should we consider:
- A simpler `selected` boolean prop on `Table.Row`?
- Having the row automatically detect selection from its `CheckCell` child?

## Testing

1. Visit `/components/table/` on the docs site
2. Click checkboxes - they should toggle once per click
3. Click the header checkbox - should select/deselect all rows
4. Selected rows should show background highlight
5. Partially selected tables should show indeterminate (minus) icon in header
6. Click the three-dots button - dropdown menu should appear

### DIFF
--- a/.changeset/clipboard-text-icon.md
+++ b/.changeset/clipboard-text-icon.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Changed ClipboardIcon to CopyIcon in ClipboardText component
--- a/.changeset/fix-kumo-add-blocks.md
+++ b/.changeset/fix-kumo-add-blocks.md
@@ -1,0 +1,7 @@
+---
+"@cloudflare/kumo": patch
+---
+
+fix(cli): include block source files in build for `kumo add` command
+
+The `kumo add` command was failing because block source files (`.tsx`) were not being copied to `dist/` during the build process. This adds copying of block source files from `src/blocks/` to `dist/src/blocks/` so the CLI can install them into user projects.
--- a/.changeset/perky-snails-stay.md
+++ b/.changeset/perky-snails-stay.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+fix: delete-resource shouldn't nest buttons within each other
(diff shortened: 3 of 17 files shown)


packages/kumo/src/components/table/table.tsx
```diff
@@ -136,73 +136,71 @@ const TableCheckCell = forwardRef<
...
+      >
+        <Checkbox
+          checked={checked}
+          indeterminate={indeterminate}
+          onCheckedChange={(newChecked) => {
+            onValueChange?.(newChecked);
+          }}
+          aria-label={label ?? "Select row"}
+          disabled={disabled}
+          className="relative before:absolute before:-inset-3 before:content-['']"
```
mattrothenberg, line 191: CSS trick to make the tap target of the checkbox fill the entire surrounding `w-10` div.

invisal: Super cool :)

PR #20, opened 2026-02-05
fix: Dialog demo Cancel button does not close dialog

awhobbs87:
## Summary

- The Cancel button in `DialogWithActionsDemo` was a plain `<Button>` with no close behavior, so clicking it did nothing
- Wrapped it with `<Dialog.Close render={...}>` to properly close the dialog, matching the pattern used by the Delete button and X close button
- Regenerated component registry to reflect the updated demo code

### DIFF
--- a/.changeset/fix-dialog-cancel.md
+++ b/.changeset/fix-dialog-cancel.md
@@ -1,5 +1,0 @@
----
-"@cloudflare/kumo": patch
----
-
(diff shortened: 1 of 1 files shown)


awhobbs87:
> Demo change looks great! Changeset is unnecessary since we're not releasing a new version of the Kumo package with this update (it's purely a docs change).

Updated!

mattrothenberg:
Oops! Did you mean to close this PR?

Issue #40, opened 2026-02-07
[docs]  Dialog docs examples use buttons that should be `Dialog.Close`


CiscoFran10:
### Describe the issue

Some Dialog documentation examples show action buttons **Cancel** that are intended to close the dialog, but they are rendered as plain `Button` components instead of being wrapped with `Dialog.Close`.

This can lead consumers to copy incorrect patterns where dialog-closing actions are not using the proper API.

https://github.com/user-attachments/assets/06ed278c-7c07-4f0f-b993-ada5ab768c90

---

### Current behavior

Example from the docs:

```tsx
<div className="mt-8 flex justify-end gap-2">
  <Button variant="secondary">Cancel</Button>
  <Dialog.Close
    render={(props) => (
      <Button variant="destructive" {...props}>
        Delete
      </Button>
    )}
  />
</div>
```

Here, **Cancel** visually behaves like a close action but is not implemented using `Dialog.Close`.

---

### Expected behavior

All actions intended to close the dialog (Cancel, Close, Delete, etc.) should consistently use `Dialog.Close` in documentation examples.

---

### Environment

* Package: `@cloudflare/kumo`
* Version: `1.0.0`
* Browser: All
* OS: All

Issue #33, opened 2026-02-07
[docs] Installation examples using relative imports need clarification on required folder structure


CiscoFran10:
## Describe the issue

The installation documentation includes examples that rely on **relative imports**, but it does not clearly explain that these imports must be adjusted to match the consumer’s local folder structure.

Because the examples are presented as copy-paste–ready, developers may assume the paths are universal. When they are not, this can lead to import errors and a few minutes of unnecessary debugging during initial setup.

---

## Current behavior

- Docs show valid examples, but:
  - They rely on relative paths
  - There is no explicit note that paths must match the project’s directory structure
- New users may copy the examples verbatim and encounter broken imports
- Debugging is simple, but the cause is not immediately obvious

---

## Expected behavior

- Installation docs should explicitly mention that:
  - Relative import paths depend on the consumer’s folder structure
  - Paths may need to be adjusted based on where the file lives
- A short inline note or callout near the examples would be sufficient
- This would help set correct expectations during first-time setup

---

## Suggested improvement (example)

Add a brief note near the relevant examples, such as:

> **Note:** These imports use relative paths. You may need to adjust them to match your project’s folder structure.

---

## Additional context

This is a small documentation improvement, but it directly impacts onboarding and developer experience. Clarifying this upfront can prevent avoidable confusion during installation, especially for first-time Kumo users.

---

## Environment and versions

- **Kumo version:** v1.0.0  
- **Docs section:** Installation → Import Styles , https://kumo-ui.com/installation/#import-styles
- **Browser:** Chrome  
- **OS:** Windows

PR #42, opened 2026-02-07
fix: address migration feedback from shadcn/Radix to Kumo

geoquant:
## Summary

Addresses issues surfaced during a ~70-component migration from shadcn/ui + Radix to Kumo.

## Changes

**Bugs**
- **CLI `doc`/`docs`/`ls` broken** — fixed registry path from `catalog/` to `ai/`, replaced internal error message with consumer-facing text
- **Tailwind 4 `@source` missing from docs** — fixed inconsistent Usage Example on installation page, added `@source` docs to package README

**Type safety**
- **Dialog DTS leaks `@base-ui/react`** — wrapped sub-components (Root, Trigger, Title, Description, Close) in thin functions using `ComponentPropsWithoutRef`, matching Popover's pattern
- **Button `aria-label` on icon-only** — discriminated union requires `aria-label` when `shape="square"` or `shape="circle"`

**Accessibility**
- **Label rendered `<span>`** — changed to `<label>` with `htmlFor` support

**Discoverability**
- `Textarea` alias for `InputArea`
- `ToastProvider` alias for `Toasty`

**Documentation**
- Added confirmation dialog recipe showing `disablePointerDismissal`
- Updated Select basic example to use `items` prop
- Documented icon-only button pattern with `aria-label`

## Not addressed

- **Dialog.Close render prop verbosity** — intentional Base UI design decision, not a Kumo concern

### DIFF
--- a/packages/kumo-docs-astro/src/lib/vite-plugin-kumo-colors.ts
+++ b/packages/kumo-docs-astro/src/lib/vite-plugin-kumo-colors.ts
@@ -130,7 +130,7 @@ export const kumoColors = ${JSON.stringify(colors, null, 2)};
       };
       moduleGraph: {
         getModuleById: (id: string) => unknown;
-        invalidateModule: (mod: unknown) => void;
+        invalidateModule: (mod: any) => void;
       };
       ws: { send: (message: { type: string }) => void };
     }) {
--- a/packages/kumo-docs-astro/src/lib/vite-plugin-kumo-registry.ts
+++ b/packages/kumo-docs-astro/src/lib/vite-plugin-kumo-registry.ts
@@ -51,7 +51,7 @@ export const kumoRegistryJson = ${json};
       };
       moduleGraph: {
         getModuleById: (id: string) => unknown;
-        invalidateModule: (mod: unknown) => void;
+        invalidateModule: (mod: any) => void;
       };
       ws: { send: (message: { type: string }) => void };
     }) {


packages/kumo-docs-astro/src/components/demos/DialogDemo.tsx
CiscoFran10: Should also be <Dialog.Close/>, see issue: #40

packages/kumo-docs-astro/src/pages/installation.astro
```diff
@@ -236,7 +236,8 @@ import "@cloudflare/kumo/styles/standalone";`}
           CSS File (app.css)
         </p>
         <CodeBlock
-          code={`@import "tailwindcss";
+          code={`@source "../node_modules/@cloudflare/kumo/dist/**/*.{js,jsx,ts,tsx}";
```
CiscoFran10, line 250: The @source "../node_modules/@cloudflare/kumo/dist/**/*.{js,jsx,ts,tsx}" example assumes a specific project structure.

It might be worth adding a short disclaimer noting that this path is relative to the user’s config file and may need adjustment. For example, if the app lives under /src, the relative path would differ (e.g. ../../node_modules/...), and copying the snippet verbatim could fail.

see issue #33

packages/kumo/README.md
```diff
@@ -44,15 +44,19 @@ import { Button } from "@cloudflare/kumo/components/button";
 
 #### For Tailwind CSS Users
 
-```js
-// Explicit import (recommended)
-import "@cloudflare/kumo/styles/tailwind";
+**Important:** Tailwind CSS v4 does not scan `node_modules/` by default. You must add a `@source` directive so Tailwind can discover the utility classes used by Kumo components. Without this, components may render with missing styles (e.g. Dialogs not centered).
```
CiscoFran10, line 47: Would it also make sense to add this on the docs installation page?

packages/kumo/ai/component-registry.json
ascorbic, line 1: Did this mean to remove all the props?

packages/kumo-docs-astro/src/pages/components/dialog.astro
```diff
@@ -123,6 +123,50 @@ export default function Example() {
         </ComponentExample>
       </div>
 
+      <div>
+        <Heading level={3}>Confirmation Dialog (Alert)</Heading>
+        <p class="mb-3 text-sm text-kumo-strong">
+          For confirmation dialogs that should not be dismissed by clicking outside (similar to an AlertDialog),
+          use <code class="rounded bg-kumo-control px-1 py-0.5 text-xs">disablePointerDismissal</code> on <code class="rounded bg-kumo-control px-1 py-0.5 text-xs">Dialog.Root</code>.
```
ascorbic, line 130: Does `disablePointerDismissal` also disable escape key dismissal?

packages/kumo-docs-astro/src/pages/installation.astro
```diff
@@ -236,7 +236,8 @@ import "@cloudflare/kumo/styles/standalone";`}
...
+@import "tailwindcss";
 @import "@cloudflare/kumo/styles/tailwind";`}
```
ascorbic: #44 says that the Kumo styles need to be imported *before* Tailwind. Which is correct? https://github.com/cloudflare/kumo/pull/44/changes#diff-aea0502ee64cb7178c22c4c0aa01d2d6dd14e33da259a2dbc934bd7a485cef8dR17-R18

packages/kumo-docs-astro/src/pages/components/dialog.astro
CiscoFran10: @ascorbic From my testing... no, disablePointerDismissal does not disable Escape key dismissal.

That actually seems like the correct behavior if we want to align with alertdialog patterns. Pressing Esc should be equivalent to activating the Cancel / Discard action and should return focus to the trigger, per the ARIA Authoring Practices:
https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/examples/alertdialog/

Also, the prop name explicitly refers to pointer dismissal, so it makes sense that it only blocks outside pointer interactions, not keyboard dismissal via Esc.

packages/kumo/ai/component-registry.json
geoquant: no, great catch! fixed

PR #46, opened 2026-02-08
fix(loader): add aria attributes for screen reader accessibility

msmps:
## Description

- Add `role="status"` and `aria-label="Loading"` to the Loader SVG component
- Resolves WCAG 2.1 SC 4.1.3 (Status Messages - Level AA) violation

## Problem

The Loader component renders an SVG spinner without any accessibility attributes, making it completely invisible to screen readers. When used as a loading indicator (e.g., in buttons, page transitions, async operations), assistive technology users receive no indication that content is loading.

Screen readers skip over the Loader entirely - users hear nothing.

## Solution

Add two ARIA attributes to the SVG element:
- `role="status"` - Indicates this is a live region for status updates
- `aria-label="Loading"` - Provides the accessible name announced by screen readers

Screen readers announce "Loading, status" when the Loader is present.

## Testing

Verified with VoiceOver on macOS:

1. Enable VoiceOver (Cmd + F5)
2. Navigate to a Loader component (Ctrl + Option + →)
3. Confirm announcement: "Loading, status"

### DIFF
--- a/.changeset/danger-token-updates.md
+++ b/.changeset/danger-token-updates.md
@@ -1,0 +1,11 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Adjust semantic tint usage for status/error ring styles across core form and feedback components.
+
+- Update background styles in `Badge` and `Banner` components to use `*-tint` tokens.
+- Update `kumo-danger` to a darker token to improve a11y contrast.
+- Update error ring styling in form components to use semantic tokens (now darker with token swap).
+- Removed `text-color-badge-red-subtle` token in replacement of the `danger` token.
+- Update theme generator color mappings used by these tints to improve visual consistency.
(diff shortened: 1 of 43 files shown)


packages/kumo/src/components/loader/loader.tsx
```diff
@@ -86,6 +86,8 @@ export const Loader = ({
       stroke="currentColor"
       className={className}
       style={{ height: sizeValue, width: sizeValue }}
+      role="status"
+      aria-label="Loading"
```
mattrothenberg: Can we make this a prop with a default value of "loading" so that we can ultimately pass a translated string down where we consume this?

mattrothenberg (changes requested):

PR #50, opened 2026-02-08
fix(a11y): add accessible labels to icon-only controls

msmps:
## Description

- Add aria-labels to icon-only controls in `MenuBar` and `Combobox`

## Problem

Several icon-only controls were missing explicit accessible labels, so screen readers could announce them as unlabeled buttons.

Controls:
- `MenuBar` option buttons (icon-only)
- `Combobox.TriggerInput` clear button
- `Combobox.TriggerInput` dropdown trigger button
- `Combobox.ChipRemove` button in multi-select chips

## Solution

Add `aria-label` to each icon-only control:

- `MenuBar`: Uses the `tooltip` prop as the aria-label (consumer-provided)
- `Combobox.TriggerInput`: New `clearLabel` and `showOptionsLabel` props with English defaults
- `Combobox.ChipRemove`: New `removeLabel` prop with English default

All labels are customizable for i18n.

---

- Reviews
  - [x] automated review not possible because: accessibility changes require manual screen reader testing
- Tests
  - [ ] Tests included/updated
  - [x] Automated tests not possible - manual testing has been completed as follows: screen reader testing to verify labels are announced

### DIFF
--- a/.changeset/odd-states-attend.md
+++ b/.changeset/odd-states-attend.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+fix(a11y): add accessible labels to icon-only controls
--- a/packages/kumo/src/components/combobox/combobox.tsx
+++ b/packages/kumo/src/components/combobox/combobox.tsx
@@ -331,6 +331,7 @@ function TriggerInput(props: ComboboxBase.Input.Props) {
       />
 
       <ComboboxBase.Clear
+        aria-label="Clear selection"
         className={cn(
           "absolute top-1/2 flex -translate-y-1/2 cursor-pointer bg-transparent p-0",
           "data-[disabled]:pointer-events-none data-[disabled]:opacity-0",
@@ -341,6 +342,7 @@ function TriggerInput(props: ComboboxBase.Input.Props) {
       </ComboboxBase.Clear>
 
       <ComboboxBase.Trigger
+        aria-label="Show options"
         className={cn(
           "absolute top-1/2 -translate-y-1/2 flex items-center justify-center cursor-pointer text-kumo-subtle",
           "m-0 bg-transparent p-0", // Reset Stratus global button styles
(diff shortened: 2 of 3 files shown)


mattrothenberg (changes requested):
Can we make these props with a default values of the english text so that we can ultimately pass a translated string down where we consume this?

geoquant (dismissed):
Thanks for the contribution — the a11y gaps you identified are real and the fix direction is correct. Merge conflict is now resolved. A few changes needed before this can land:

## 1. Make labels configurable props (per @mattrothenberg's feedback)

Hardcoded English strings prevent i18n. Each label should be a prop with an English default.

**`TriggerInput`:**
```tsx
function TriggerInput({
  clearLabel = "Clear selection",
  showOptionsLabel = "Show options",
  ...props
}: ComboboxBase.Input.Props & {
  /** Accessible label for the clear button. @default "Clear selection" */
  clearLabel?: string;
  /** Accessible label for the dropdown trigger. @default "Show options" */
  showOptionsLabel?: string;
}) {
  // ...
  <ComboboxBase.Clear aria-label={clearLabel} ...>
  <ComboboxBase.Trigger aria-label={showOptionsLabel} ...>
}
```

**`Chip`:**
```tsx
function Chip({
  removeLabel = "Remove",
  ...props
}: ComboboxBase.Chip.Props & {
  /** Accessible label for the chip remove button. @default "Remove" */
  removeLabel?: string;
}) {
  // ...
  <ComboboxBase.ChipRemove aria-label={removeLabel} ...>
}
```

## 2. Chip label wording

Change `"Remove selection"` → `"Remove"`. In chip context, the chip *is* the selection — "Remove selection" is ambiguous when there are multiple chips.

## 3. MenuBar — looks good as-is

`aria-label={tooltip}` on the icon-only button is correct and necessary — the Tooltip component provides `aria-describedby` (supplementary), not `aria-label` (name). No change needed here.

stritt (approved):
looks good, thanks for making the labels configurable 👍

mattrothenberg (approved):

PR #72, opened 2026-02-10
fix(docs): fix code block border overflow in ComponentExample

mattrothenberg:
## Summary

- Fixes visual overflow artifact in component example code snippets
- Uses more specific CSS selector targeting `.code-block` directly instead of generic `.rounded-lg`

Peep the bottom left and right corners.

**Before**
Screenshot 2026-02-10 at 2 55 24 PM

**After**
Screenshot 2026-02-10 at 2 55 13 PM


packages/kumo-docs-astro/src/components/SearchDialog.tsx
superhighfives, line 248: @mattrothenberg should these also be in here?

mattrothenberg: fixed

PR #79, opened 2026-02-11
feat(pagination): add ability to overwrite pagination text

dochne:
## What have you changed?
This adds support for a `text` argument for the Pagination component. I've done this by adding a `text` argument which accepts a method that can return `React.ReactNode`. It takes a prop argument of all the variables that a user might want to use to generate custom text.

## What are you trying to achieve?
I have a table of data which can have thousands of entries in it. Pagination is done via token.

In current usage, we are aware of what the totalCount is, which allows me to pass it in and this will all work fine.

However, we are looking at adding search functionality to this table. In this new circumstance, we will *not* be aware of what the total count is. Two things are true of our current API design:
1. in order to have a next button enabled, totalCount needs to be set
2. if we have totalCount set, it will appear in this text

While the wrapping interface will be a little ugly, this will allow me to "fake" the totalCount if I have a continuation token - while ensuring that this isn't shown in the text.

## What other possibilities are there?
I've explored this space a bit in my previous PR, which introduced "simple" pagination for this component. Even if this is a little gross on my side, this feels the cleanest way from a this API point of view.

Other possible iterations on the ```text={({perPage}) => `Hey! ${perPage}`}``` solution could be either:
- just allow text ```text={`Hey! ${perPage}`}``` - as all of the context is technically being passed into the Pagination component, you could avoid the method. I like the fact that our current method gives us easy access to the ```{pageShowingRange}``` variable however
- allow this in addition to the callback (e.g. support ```text={"my text"}``` and ```text={() => "my text"}```
- instead of passing {pageShowingRange} as a single variable that looks like 1-9, we could pass a ```{lowerItem}``` and ```{upperItem}``` instead (or similar), as the dash feels a bit like a presentation concern

Other potential changes here could be that `text` isn't a great prop name - `description` could work better? Happy to change if there are any better suggestions

### DIFF
--- a/.changeset/clipboard-text-icon.md
+++ b/.changeset/clipboard-text-icon.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Changed ClipboardIcon to CopyIcon in ClipboardText component
--- a/.github/workflows/bonk.yml
+++ b/.github/workflows/bonk.yml
@@ -1,0 +1,38 @@
+name: Bonk
+
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.issue.number || github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  bonk:
+    if: github.event.sender.type != 'Bot'
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: write
+      issues: write
+      pull-requests: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Run Bonk
+        uses: ask-bonk/ask-bonk/github@main
+        env:
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CF_AI_GATEWAY_ACCOUNT_ID }}
+          CLOUDFLARE_GATEWAY_ID: ${{ secrets.CF_AI_GATEWAY_NAME }}
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CF_AI_GATEWAY_TOKEN }}
+        with:
+          model: "cloudflare-ai-gateway/anthropic/claude-opus-4-6"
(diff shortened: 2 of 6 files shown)


packages/kumo-docs-astro/src/components/demos/PaginationDemo.tsx
```diff
@@ -52,3 +52,16 @@ export function PaginationLargeDatasetDemo() {
     <Pagination page={page} setPage={setPage} perPage={25} totalCount={1250} />
   );
 }
+
+export function PaginationCustomTextDemo() {
+  const [page, setPage] = useState(1);
+  return (
+    <Pagination
+      text={({ perPage }) => `Page ${page} - showing ${perPage} per page`}
+      page={1}
```
CiscoFran10: Needs to pass the current `page` state as a prop instead of a static value:

```tsx
<Pagination
  page={page}
/>
```

packages/kumo-docs-astro/src/pages/components/pagination.astro
```diff
@@ -147,6 +148,24 @@ export default function Example() {
...
+
+      <div>
+        <Heading level={3}>Custom Text</Heading>
+        <p class="mb-4 text-kumo-subtle">
+          You can set custom pagination text
+        </p>
+        <ComponentExample
+          code={`<Pagination
+  text={({perPage}) => \`Showing $\{perPage\} per page!\`}
+  page={1}
```
CiscoFran10, line 160: Same here

packages/kumo/ai/schemas.ts
```diff
@@ -583,6 +583,7 @@ export const PaginationPropsSchema = z.object({
   page: z.number().optional(), // Current page number (1-indexed).
   perPage: z.number().optional(), // Number of items displayed per page.
   totalCount: z.number().optional(), // Total number of items across all pages.
+  text: z.unknown().optional(),
```
CiscoFran10: I think this file is auto-generated, but shouldn’t this define the correct parameter and return types instead of `z.unknown()`?

dochne: It looks like the autogeneration just does `z.unknown()` for a lot of things for now - even if typing it properly is possible, it'll just get overwritten

PR #87, opened 2026-02-13
fix(MenuBar): Updated the MenuBar so its child buttons align with the container’s outer corners.

### LINKED ISSUE #52: MenuBar children have wrong border radius
Image

rodrigosdev:
Updated the MenuBar so its child buttons align with the container’s outer corners.
  Previously all buttons used `rounded-md` (6px) while the container uses `rounded-lg` (8px), which made the first
  and last buttons look cut off.
  They now use `first:rounded-l-lg` `last:rounded-r-lg`, so:
  • First button: left corners match the container’s 8px radius
  • Last button: right corners match the container’s 8px radius
  • Middle buttons: keep `rounded-md` on all sides
  • Single button: gets `rounded-l-lg` and `rounded-r-lg`, effectively matching `rounded-lg`

fixes #52


.changeset/common-clowns-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/kumo": major
```
rozenmd:
```suggestion
"@cloudflare/kumo": patch
```

PR #90, opened 2026-02-13
feat(clipboard-text): add slide animation with tooltip on copy

najlaskr:
## Summary
The existing `ClipboardText` component lacks visual confirmation upon clicking, leading to potential double-clicking. This PR adds a micro-interaction and tooltip to confirm the action is complete. 

**Proposed Change**

https://github.com/user-attachments/assets/47f4a864-0ddf-40fb-a845-4cc5b5fad0f1

## Key Changes
- Immediate feedback loop: added an animated transition from the "copy" icon to a "checkmark" icon
- Contextual validation: integrated a "copied" tooltip for additional confirmation
- Reduced the animation time from 2000ms to 1100ms to avoid the "unresponsive" feel of the UI

### DIFF
--- a/packages/kumo/src/components/clipboard-text/clipboard-text.tsx
+++ b/packages/kumo/src/components/clipboard-text/clipboard-text.tsx
@@ -214,9 +214,10 @@ export const ClipboardText = forwardRef<HTMLDivElement, ClipboardTextProps>(
           }
         }
 
+        setCopied(true);
+
         // Show anchored toast if tooltip mode is enabled
         if (tooltip) {
-          setCopied(true);
           clipboardToastManager.add({
             description: copiedText,
             positionerProps: {
@@ -229,6 +230,9 @@ export const ClipboardText = forwardRef<HTMLDivElement, ClipboardTextProps>(
               setCopied(false);
             },
           });
+        } else {
+          // Reset copied state after delay when no tooltip
+          setTimeout(() => setCopied(false), 1500);
         }
 
         onCopy?.();


packages/kumo/src/components/clipboard-text/clipboard-text.tsx
```diff
@@ -161,16 +175,42 @@ export const ClipboardText = forwardRef<HTMLDivElement, ClipboardTextProps>(
...
+          open={showTooltip && copied}
+          asChild
         >
-          {copied ? <CheckIcon /> : <CopyIcon />}
+        <Button
+            size={sizeConfig.buttonSize}
+            variant="ghost"
+            className="rounded-none border-l! border-kumo-line! px-3 relative overflow-hidden transition-all duration-200"
+            onClick={copyToClipboard}
+            aria-label={copied ? "Copied" : "Copy to clipboard"}
```
CiscoFran10: When `copied` is `true`, the button’s `aria-label` changes to “Copied”, which reflects a result rather than the action the control performs. Since accessible names should describe the control’s purpose, it would be better to keep a stable label “Copy to clipboard”.

The success feedback (“Copied to clipboard”) can instead be announced via the existing `aria-live` region on line 214.

packages/kumo/src/components/clipboard-text/clipboard-text.tsx
```diff
@@ -161,16 +175,42 @@ export const ClipboardText = forwardRef<HTMLDivElement, ClipboardTextProps>(
...
-        <Button
-          size={sizeConfig.buttonSize}
-          variant="ghost"
-          className="rounded-none border-l! border-kumo-line! px-3"
-          onClick={copyToClipboard}
-          aria-label={copied ? "Copied" : "Copy to clipboard"}
-          aria-pressed={copied}
+        <Tooltip
+          content="Copied"
+          side="bottom"
```
CiscoFran10: **Optional:** Consider exposing the tooltip `side` as a prop so consumers can override the default (`"bottom"`).

Right now it’s hardcoded:

```tsx
<Tooltip
  content="Copied"
  side="bottom"
/>
```

packages/kumo/src/components/clipboard-text/clipboard-text.tsx
mattrothenberg: **A11y issue:** `aria-label` should remain stable — it describes what the button *does*, not its current state.

Changing to "Copied" is confusing for screen reader users since the button still performs the copy action when clicked again. The success feedback is already announced via the `aria-live` region on line 214.

```suggestion
            aria-label="Copy to clipboard"
```

Also, `aria-pressed` should be removed — this is an action button with transient feedback, not a toggle. Toggle buttons have two stable states (on/off); this button always does the same thing (copy).

packages/kumo/src/components/clipboard-text/clipboard-text.tsx
```diff
@@ -161,16 +175,42 @@ export const ClipboardText = forwardRef<HTMLDivElement, ClipboardTextProps>(
...
       >
         <span className="grow px-4">{text}</span>
-        <Button
-          size={sizeConfig.buttonSize}
-          variant="ghost"
-          className="rounded-none border-l! border-kumo-line! px-3"
-          onClick={copyToClipboard}
-          aria-label={copied ? "Copied" : "Copy to clipboard"}
-          aria-pressed={copied}
+        <Tooltip
```
mattrothenberg: **i18n blocker:** Hardcoded "Copied" string cannot be translated.

Suggested approach — expose tooltip config:
```tsx
tooltip?: false | {
  content?: string;  // default: "Copied"
  side?: "top" | "bottom" | "left" | "right";
};
```

This also addresses @CiscoFran10's feedback about the hardcoded `side="bottom"`.

packages/kumo/src/components/clipboard-text/clipboard-text.tsx
```diff
@@ -74,6 +85,8 @@ export interface ClipboardTextProps extends KumoClipboardTextVariantsProps {
   className?: string;
```
mattrothenberg, line 88: **i18n blocker:** The `showTooltip` prop is too narrow. Consider this API instead:

```tsx
/** Tooltip config. Pass false to disable. @default false */
tooltip?: false | {
  content?: string;
  side?: "top" | "bottom" | "left" | "right";
};

/** Accessible labels for i18n. */
labels?: {
  /** @default "Copy to clipboard" */
  copyAction?: string;
  /** @default "Copied to clipboard" */  
  copySuccess?: string;
};
```

This consolidates i18n concerns and makes the tooltip opt-in (the icon animation is sufficient feedback for most cases).

packages/kumo/src/components/clipboard-text/clipboard-text.tsx
```diff
@@ -29,6 +30,16 @@ export const KUMO_CLIPBOARD_TEXT_DEFAULT_VARIANTS = {
   size: "lg",
 } as const;
```
mattrothenberg, line 35: **Nitpick:** Let's simplify this for clarity. The string concatenation in `initial` and `end` makes it hard to see the full class list at a glance.

Consider inlining or using a clearer pattern:

```tsx
const iconStyles = {
  base: "absolute inset-0 flex items-center justify-center transition-all duration-200",
  visible: "translate-y-0 opacity-100",
  above: "-translate-y-full opacity-0 pointer-events-none",
  below: "translate-y-full opacity-0 pointer-events-none",
} as const;
```

packages/kumo/src/components/clipboard-text/clipboard-text.tsx
```diff
@@ -161,18 +222,33 @@ export const ClipboardText = forwardRef<HTMLDivElement, ClipboardTextProps>(
...
-          size={sizeConfig.buttonSize}
-          variant="ghost"
-          className="rounded-none border-l! border-kumo-line! px-3"
-          onClick={copyToClipboard}
-          aria-label={copied ? "Copied" : "Copy to clipboard"}
-          aria-pressed={copied}
-        >
-          {copied ? <CheckIcon /> : <CopyIcon />}
-        </Button>
+        {tooltip ? (
```
CiscoFran10, line 286: I believe the correct pattern should be: [Anchored toasts](https://base-ui.com/react/components/toast#anchored-toasts).

Base UI recommends using Toast for this type of feedback.

mattrothenberg: There is an open issue on this question and it hasn't been resolved yet. I'd prefer if we kept our current approach for right now: https://github.com/mui/base-ui/issues/4113

packages/kumo/src/components/clipboard-text/clipboard-text.tsx
```diff
@@ -91,66 +115,103 @@ export const ClipboardText = forwardRef<HTMLDivElement, ClipboardTextProps>(
...
       }
     }, [copied]);
 
+    const copyButton = (
+      <Button
+        size={sizeConfig.buttonSize}
+        variant="ghost"
+        className="rounded-none border-l! border-kumo-line! px-3 relative overflow-hidden transition-all duration-200"
+        onClick={copyToClipboard}
+        aria-label={labels.copyAction ?? "Copy to clipboard"}
```
CiscoFran10: Shouldn’t the defaults be handled in the props destructuring on line 119 instead?

packages/kumo/src/components/clipboard-text/clipboard-text.tsx
```diff
@@ -161,18 +222,33 @@ export const ClipboardText = forwardRef<HTMLDivElement, ClipboardTextProps>(
...
+            >
+              {tooltip.content(copied)}
+            </Popover.Content>
+          </Popover>
+        ) : (
+          copyButton
+        )}
         <span className="sr-only" aria-live="polite">
-          {copied ? "Copied to clipboard" : ""}
+          {copied ? (labels.copySuccess ?? "Copied to clipboard") : ""}
```
CiscoFran10: > Shouldn’t the defaults be handled in the props destructuring on line 119 instead?

Same here

packages/kumo/src/components/clipboard-text/clipboard-text.tsx
CiscoFran10: They do have a working example in their docs though, for copy buttons:

https://stackblitz.com/edit/dst7ywbd?file=src%2FApp.tsx

CiscoFran10: But its okay, I’ll keep tracking that issue, and if anything changes, I’ll let you know.

mattrothenberg: 8498a02e

mattrothenberg (approved):

PR #106, opened 2026-02-17
fix(docs): Theme toggle scrollbar overlap and scroll issues

CiscoFran10:
The theme toggle was positioned absolutely in the top-right corner, which caused it to be cut off by the scrollbar on Windows. Additionally, there were scroll issues at certain breakpoints where unwanted scrollbars appeared.

image

---

image 

### 1. Moved theme toggle into header components

Instead of using absolute positioning (`fixed top-0 right-2`), the theme toggle is now integrated directly into the header components:

- **Header.astro**: Added theme toggle to the desktop header with proper flexbox layout
- **StickyDocHeader.tsx**: Integrated theme toggle into the sticky header on documentation pages
- **SidebarNav.tsx**: Added theme toggle to the mobile header (replacing the spacer div)

### 2. Restructured header layout

Changed the header from using absolute positioning and margins to a flexbox-based layout:
- Header now uses `flex` with proper `min-w-0` and `flex-1` utilities to handle overflow correctly
- Theme toggle container uses `w-12 shrink-0` to maintain consistent width and prevent shrinking
- Removed `md:pr-12` padding that was attempting to compensate for the absolute-positioned toggle

### 3. Fixed scrollbar issues

Added `overflow: clip` to the body element in `global.css` to prevent unwanted scrollbars at various breakpoints. This ensures that only the intended scrollable containers (like the main content area) show scrollbars.

### Related

#18
#15


packages/kumo-docs-astro/src/components/docs/StickyDocHeader.tsx
```diff
@@ -1,6 +1,7 @@
 import { useEffect, useState } from "react";
 import { cn } from "@cloudflare/kumo";
-import { GithubLogo } from "@phosphor-icons/react";
+import { GithubLogoIcon } from "@phosphor-icons/react";
```
CiscoFran10, line 3: Small fix: `GithubLogo` is deprecated, phosphor recommends  `GithubLogoIcon`, lmk if this should be a separate PR.

packages/kumo-docs-astro/src/styles/global.css
```diff
@@ -21,3 +21,7 @@
 html {
   font-display: optional;
 }
+
+body {
+  overflow: clip;
```
CiscoFran10: Why `clip` instead of `hidden` ??  Guarantees the element is not a scroll container at all, whereas overflow: hidden still creates a scroll container internally, allowing programmatic scrolling and participation in scrollbar layout mechanics.

Browser support: Baseline, Widely available across major browsers, Global 94.86%

Issue #38, opened 2026-02-07
[breadcrumbs] Breadcrumbs are hidden on mobile viewport


CiscoFran10:
## Describe the issue

On the **Breadcrumbs documentation page**, the breadcrumbs component disappears entirely on mobile / small viewports.

<img width="351" height="861" alt="Image" src="https://github.com/user-attachments/assets/cdff0104-32dc-4b05-9edb-f64cdd84e62c" />

From inspecting the implementation, this appears to be caused by the root container using responsive utility classes that hide the component below the `sm` breakpoint:

```ts
export function breadcrumbsVariants({
  size = KUMO_BREADCRUMBS_DEFAULT_VARIANTS.size,
}: KumoBreadcrumbsVariantsProps = {}) {
  return cn(
    "group mr-4 hidden min-w-0 grow items-center sm:flex",
    KUMO_BREADCRUMBS_VARIANTS.size[size].classes,
  );
}
````

Because of `hidden sm:flex`, breadcrumbs are never rendered on mobile.

If this behavior is **intentional**, it would be helpful to document the reasoning and recommended mobile navigation alternatives.
If it is **unintentional**, this prevents breadcrumbs from being evaluated or used in mobile layouts.

---

## Expected behavior

One of the following should be true:

* Breadcrumbs remain visible on mobile (possibly with truncation or overflow handling), **or**
* The docs clearly explain that breadcrumbs are intentionally hidden on mobile and why

---

## Actual behavior

* Breadcrumbs are completely hidden on mobile screens
* No explanation or guidance is provided in the documentation
* This can be confusing when evaluating the component’s behavior across breakpoints

---

## Environment

* **Package:** @cloudflare/kumo
* **Component:** Breadcrumbs
* **Docs site:** [https://kumo-ui.com/](https://kumo-ui.com)
* **Browser:** Chrome
* **Viewport:** Mobile / small screens

PR #111, opened 2026-02-17
fix(breadcrumbs): make breadcrumbs visible on mobile viewports

rodrigosdev:
Removed `hidden sm:flex` from the breadcrumbs container to ensure visibility across all screen sizes. This change addresses the issue where breadcrumbs were completely hidden below the `sm` breakpoint, hindering their use in mobile layouts. See issue #38 for more details.

### DIFF
--- a/.changeset/deprecate-date-range-picker.md
+++ b/.changeset/deprecate-date-range-picker.md
@@ -1,8 +1,0 @@
----
-"@cloudflare/kumo": patch
----
-
-Deprecate DateRangePicker in favor of DatePicker with `mode="range"`
-
-- Added `@deprecated` JSDoc comments to DateRangePicker component and its exports
-- IDEs will now show deprecation warnings guiding users to use DatePicker instead
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,273 @@
+# Contributing
+
+Kumo is an open-source project and we welcome contributions from you. Thank you!
+
+Below you can find some guidance on how to be most effective when contributing to the project.
+
+## tl;dr for contributing to Kumo
+
+Useful commands for developing Kumo (all commands below should be run in the project root):
+
+- `pnpm i; pnpm build` will build everything in kumo.
+- `pnpm dev` will watch and build changes while you develop.
+
+Before committing/submitting a PR:
+
+- Add a [changeset](#changesets) with `pnpm changeset`.
+- Don’t squash your commits after a review.
+
+## Before getting started
+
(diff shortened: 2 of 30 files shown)


packages/kumo/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -113,6 +125,14 @@ function Separator() {
   );
 }
 
+function MobileEllipsis() {
```
CiscoFran10, line 128: Should this be restricted to mobile only? I can think of scenarios where we would want to shows ellipsis on desktop as well, such as overflow menus in breadcrumbs or constrained navigation layouts.. se reference:

https://ui.shadcn.com/docs/components/base/breadcrumb

Maybe it makes sense to generalize this into `Breadcrumb.Ellipsis` component and export as one of the parts?

packages/kumo/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -192,16 +212,56 @@ export function Breadcrumb({
   size = "base",
   className,
 }: BreadcrumbsProps) {
+  const childArray = Children.toArray(children);
+  const mobileChildren = getMobileBreadcrumbChildren(childArray);
```
CiscoFran10, line 216: This should probably be optional rather than always applied. There may be valid use cases where the full breadcrumb should remain visible across all breakpoints.

Consider adding an explicit prop to control this, for example:

```tsx
<Breadcrumb collapse="mobile" />
```

or

```tsx
<Breadcrumb enableMobileCollapse />
```

packages/kumo/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -192,16 +212,56 @@ export function Breadcrumb({
...
 }
 
+function isComponentElement(
+  child: ReactNode,
+  component: unknown,
+): child is ReactElement {
+  return isValidElement(child) && child.type === component;
+}
+
+function getMobileBreadcrumbChildren(children: ReactNode[]): ReactNode[] {
```
CiscoFran10, line 236: Instead of relying on inspecting `children` to figure out which items should collapse, it might be cleaner to expose an `items` prop and render the breadcrumb from a structured array.

That way the collapse logic becomes straightforward and avoids needing checks like `child.type === Link` or cloning elements. It also makes it easier to support things like ellipsis menus, dynamic breadcrumbs from routing, or marking specific items as collapsible.

Something like:

```tsx
<Breadcrumb
  items={[
    { label: "Home", href: "/" },
    {
      label: "Components",
      href: "/components",
      collapsible: true,
    },
    {
      label: "Navigation",
      menu: [
        { label: "Breadcrumb", onSelect: () => {} },
        { label: "Tabs", onSelect: () => {} },
      ],
    },
    { label: "Breadcrumb", current: true },
  ]}
  collapse="mobile"
/>
```

PR #114, opened 2026-02-18
fix(select): add overflow handling and truncation for long trigger text

invisal:
The Select component's trigger button could overflow its container when labels were long, causing layout issues. This fix adds `min-w-0` and `overflow-hidden` to the trigger wrapper and `truncate` to the value element so long labels are clipped cleanly within the button bounds.

Demo examples have been updated to include long-label and multi-line scenarios to document and verify the behavior.

image

image

### DIFF
--- a/packages/kumo/src/components/select/select.tsx
+++ b/packages/kumo/src/components/select/select.tsx
@@ -263,7 +263,7 @@ export function Select<T, Multiple extends boolean | undefined = false>({
         {loading ? (
           <SkeletonLine className="w-32" />
         ) : (
-          <SelectBase.Value className="min-w-0 flex-1 truncate text-left">
+          <SelectBase.Value className="min-w-0 flex-1 truncate text-start">
             {renderValue}
           </SelectBase.Value>
         )}


packages/kumo/src/components/select/select.tsx
```diff
@@ -227,7 +227,9 @@ export function Select<T, Multiple extends boolean | undefined = false>({
         {loading ? (
           <SkeletonLine className="w-32" />
         ) : (
-          <SelectBase.Value>{renderValue}</SelectBase.Value>
+          <SelectBase.Value className="min-w-0 flex-1 truncate text-left">
```
mattrothenberg: This should use `text-start` so that we can support RTL languages.

invisal: Good catch.

geoquant:
do you mind adding these oxlint changes to your pr, i think this will help you pass lint:
https://github.com/cloudflare/kumo/pull/132/changes#diff-c5364e548d3fc49147095fce945983b0b4741304887e14089cabbfb9a15a3bb2

i had that PR open but it was doing too many things and not getting reviews

otherwise, we run the risk of introducing too modern api's for our dashboard

PR #138, opened 2026-02-22
fix(docs): theme toggle position

vaibhavshn:
Fixes the theme toggle position when in desktop mode and the main content container overflows and shows a scrollbar. It displaced the fixed theme toggle position. This PR keeps the mobile position as is, only changes position in desktop view.

| Before | After |
|--|--|
| Screenshot 2026-02-22 at 15 09 41@2x | Screenshot 2026-02-22 at 15 10 13@2x |

### DIFF
--- a/packages/kumo-docs-astro/src/layouts/MainLayout.astro
+++ b/packages/kumo-docs-astro/src/layouts/MainLayout.astro
@@ -41,7 +41,7 @@ const currentPath = Astro.url.pathname;
 
     <!-- Theme toggle: fixed in top right corner -->
     <div
-      class="pointer-events-auto fixed top-0 right-2 z-50 flex h-[49px] items-center"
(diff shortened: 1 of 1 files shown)


packages/kumo-docs-astro/src/components/Header.astro
```diff
@@ -7,6 +9,9 @@ const kumoVersion = typeof __KUMO_VERSION__ !== "undefined" ? __KUMO_VERSION__ :
 <header
   class="sticky top-0 z-10 border-b border-kumo-line bg-kumo-elevated md:pr-12"
 >
+  <div class="absolute top-0 right-0 hidden h-12 w-12 items-center justify-center md:flex">
```
CiscoFran10: This Header is only used in the Home page. for the other pages Theme toggle is missing.

Image

vaibhavshn: Ah nice catch. Will check.

PR #142, opened 2026-02-23
feat(pagination): add optional page size selector

lkipke:
#  Description

Adds an optional page size selector to the Pagination component.

Screenshot 2026-02-23 at 12 53 32 PM

### DIFF
--- a/packages/kumo/src/components/pagination/pagination.tsx
+++ b/packages/kumo/src/components/pagination/pagination.tsx
@@ -108,7 +108,12 @@ function PaginationInfo({ children, className }: PaginationInfoProps) {
       : null;
 
   return (
-    <div className={cn("text-sm text-kumo-strong", className)}>{content}</div>
+    <div
+      data-slot="pagination-info"
+      className={cn("text-sm text-kumo-strong", className)}
+    >
+      {content}
+    </div>
   );
 }
 
@@ -139,7 +144,10 @@ function PaginationPageSize({
   className,
 }: PaginationPageSizeProps) {
   return (
-    <div className={cn("flex items-center gap-2", className)}>
+    <div
+      data-slot="pagination-page-size"
+      className={cn("flex items-center gap-2", className)}
+    >
       {label && <span className="text-sm text-kumo-strong">{label}</span>}
       <Select
         label="Page size"
@@ -175,7 +183,10 @@ function PaginationControls({
     usePaginationContext();
 
   return (
-    <div className={cn("grow flex flex-col items-end", className)}>
+    <div
+      data-slot="pagination-controls"
+      className={cn("grow flex flex-col items-end", className)}
+    >
       <div>
         <InputGroup focusMode="individual">
           {controls === "full" && (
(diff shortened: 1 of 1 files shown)


packages/kumo/src/components/pagination/pagination.tsx
```diff
@@ -120,9 +148,38 @@ export function Pagination({
...
   return (
     <div className="flex items-center justify-between gap-2">
       <div className="grow text-sm text-kumo-strong">{getPaginationText()}</div>
+      {showPageSizeSelector && (
+        <>
+          <div className="border-l border-kumo-line mx-2 h-6" />
+          <div className="flex items-center gap-2">
+            <span className="text-sm text-kumo-strong">
+              {getPageSizeText()}
+            </span>
```
lkipke: Wasn't sure if we wanted a separate prop for this, or if we instead wanted to only provide the one `text` prop and omit this if it's provided

packages/kumo/src/components/pagination/pagination.tsx
```diff
@@ -197,3 +250,261 @@ export function Pagination({
...
+      return text({ page, perPage, totalCount, pageShowingRange });
+    } else if (totalCount && totalCount > 0) {
+      return `Showing ${pageShowingRange} of ${totalCount}`;
+    }
+    return null;
+  };
+
+  return (
+    <PaginationContext.Provider value={contextValue}>
+      <div className={cn("flex items-center gap-2 w-full", className)}>
```
lkipke: This `w-full` is somewhat opinionated, combined with the default `className` on `PaginationControls`. By default, it'll fill the space with the controls on the right and everything else on the left:

Image

This is how it's used in the places that I've seen, but happy to make it less opinionated if preferred.

packages/kumo/src/components/pagination/pagination.tsx
```diff
@@ -197,3 +250,261 @@ export function Pagination({
...
+    totalCount?: number;
+    pageShowingRange: string;
+  }) => ReactNode;
+  /**
+   * @deprecated Use compound component with Pagination.PageSize instead
+   * @default false
+   */
+  showPageSizeSelector?: boolean;
+  /**
+   * @deprecated Use Pagination.PageSize options prop instead
```
lkipke: I assumed if we were introducing a new compound API, we'd want to deprecate the old one so that we don't have to maintain both.

packages/kumo/src/components/pagination/pagination.tsx
mattrothenberg: This totally makes sense to me. I wonder if it makes sense to expose a data attribute or SOMETHING on that element so that it's easy to override in user land if need be.

mattrothenberg: cc @msmps, I think this is what you were mentioning?

msmps: > cc @msmps, I think this is what you were mentioning?

that's right! i was thinking something similar to other libraries that allow you to target "slots" with an attribute like `data-slot` so you're able to override styles via `[data-slot="pagination-<slot>"]`

mattrothenberg: @lkipke Probably worth adding this. I suspect users need an escape hatch for overriding this design choice.

lkipke: Added!

PR #185, opened 2026-03-05
[kumo] Add background tint to secondary button hover state

tccloudflare:
Added the color Kumo-tint to the secondary button hover state.

### DIFF
--- a/.changeset/secondary-button-hover-tint.md
+++ b/.changeset/secondary-button-hover-tint.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Add subtle background tint on secondary button hover state for better visual feedback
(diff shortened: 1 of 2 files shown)


mattrothenberg (changes requested):
Thank you! Can we add classes for the focus state, too?

Brayden (commented):
Looking between the proposed changes here in the Docs Preview link versus what is live in kumo-ui.com but can't spot the difference on hover state. 

I think the only change here (now) is the focus state on a not-disabled secondary button. Can we change the patch notes to reflect that by chance?

stritt:
thanks for this! closing this one since the secondary button hover and focus behavior has since been covered by newer button updates. feel free to reopen if you think it still makes sense.

PR #202, opened 2026-03-10
feat(date-picker): add restartable range selection

singkia:
## Summary
- add an opt-in `rangeSelectionBehavior="restart"` mode for `DatePicker` range selection
- expose `onRangeComplete` so popover-based range pickers can close after the second click
- add regression coverage and update the docs demos to use committed vs draft range state

## Why
Popover range pickers should behave like a bounded form interaction: first click starts a draft, second click confirms the range, and the next click after completion should start a new draft instead of mutating the previously confirmed range.

## Verification
- `pnpm --dir packages/kumo exec oxlint --config .oxlintrc.json src/components/date-picker/date-picker.tsx src/components/date-picker/date-picker.test.tsx --type-aware`
- `pnpm --dir packages/kumo exec eslint src/components/date-picker/date-picker.tsx src/components/date-picker/date-picker.test.tsx src/components/date-picker/index.ts src/index.ts`
- `pnpm --filter @cloudflare/kumo exec vitest run --project=unit src/components/date-picker/date-picker.test.tsx`

## Notes
- `pnpm --filter @cloudflare/kumo-docs-astro exec astro check` still reports unrelated existing `echarts` type errors outside the DatePicker docs files.


mattrothenberg:
Thanks for this contribution. At this moment, we are going to pass on adding this functionality and we will reach out to you if we want to bring it back!

PR #204, opened 2026-03-10
fix(tabs): improve focus ring and hover styling

mattrothenberg:
## Summary

- Fixed focus ring to use proper `ring-kumo-ring` token (gray) instead of browser default blue
- Segmented variant: use inset focus ring on all tabs (including active tab) so focus is always visible when keyboard navigating
- Underline variant: add padding (`px-2 py-2.5`) so focus ring has breathing room around text
- Added subtle hover states for both variants (text color for segmented, background for underline)

## Before/After

**Before:** Browser blue focus ring, too tight on underline tabs, no visible focus on active segmented tab

**After:** Consistent gray focus ring with proper spacing, clean hover states, focus visible on all tabs

## Changes

- `packages/kumo/src/components/tabs/tabs.tsx` - Focus and hover styling updates

### DIFF
--- a/.changeset/fix-banner-contrast.md
+++ b/.changeset/fix-banner-contrast.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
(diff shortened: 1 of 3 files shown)


m0tts:
Looks and works great! Thanks for adding @mattrothenberg

PR #209, opened 2026-03-12
chore(kumo): replace package ESLint pass with Oxlint plugins

geoquant:
## Summary
- replace the package-level ESLint pass with Oxlint JS plugin coverage for the remaining `jsx-a11y` rules
- disable `unicorn/no-array-sort` and `unicorn/no-array-reverse` so Kumo does not steer consumers toward ES2023 runtime APIs that still break Stratus browser targets
- pin the library build target to `es2022` and add a post-build browser-compat test that scans `dist/` for banned newer runtime APIs

## Validation
- `pnpm lint`
- `pnpm --filter @cloudflare/kumo exec vitest run --project=unit tests/build/browser-compat.test.ts`

### DIFF
--- a/.changeset/fix-deep-imports-test.md
+++ b/.changeset/fix-deep-imports-test.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Fix flaky vitest "Closing rpc while fetch was pending" error in deep-imports test
--- a/packages/kumo-docs-astro/src/layouts/DocLayout.astro
+++ b/packages/kumo-docs-astro/src/layouts/DocLayout.astro
@@ -20,11 +20,16 @@ const baseUIUrl = baseUIComponent
   : null;
 
 // Extract the component name from the path (e.g., "components/button" -> "button")
-// and construct the direct file URL (e.g., "components/button/button.tsx")
+// and construct the direct file URL (e.g., "components/button/button.tsx").
+// If sourceFile already has a file extension, use it as-is.
 const githubSourceUrl = sourceFile
   ? (() => {
+      const base = `https://github.com/cloudflare/kumo/blob/main/packages/kumo/src/`;
+      if (/\.\w+$/.test(sourceFile)) {
+        return `${base}${sourceFile}`;
+      }
       const componentName = sourceFile.split("/").pop();
-      return `https://github.com/cloudflare/kumo/blob/main/packages/kumo/src/${sourceFile}/${componentName}.tsx`;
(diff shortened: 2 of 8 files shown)


packages/kumo/.oxlintrc.json
```diff
@@ -14,6 +20,19 @@
...
-    "jsx-a11y/no-autofocus": "off"
+    "jsx-a11y/no-autofocus": "off",
+    // These unicorn rules auto-fix .sort()/.reverse() to .toSorted()/.toReversed()
+    // (ES2023). The library still needs to avoid newer runtime Array APIs in
+    // published dist output, so keep the source rule off and let the post-build
+    // browser-compat test catch regressions instead.
+    "unicorn/no-array-sort": "off",
+    "unicorn/no-array-reverse": "off",
+    "jsx-a11y-js/aria-proptypes": "error",
+    "jsx-a11y-js/interactive-supports-focus": "error",
```
Bryant-Curtis, line 35: I haven't tried this locally so please correct me if I'm wrong, but from [this issue ](https://github.com/oxc-project/oxc/issues/1141) I'd assume this `jsx-a11y-js/interactive-supports-focus` + following 4 rules are still not supported in oxlint.

jsx-a11y/no-interactive-element-to-noninteractive-role 
jsx-a11y/no-noninteractive-element-interactions
jsx-a11y/no-noninteractive-element-to-interactive-role
jsx-a11y/control-has-associated-label

That said, the `jsx-a11y-js/aria-proptypes` and `jsx-a11y-js/no-static-element-interactions` do seem to now be supported and can be removed from the eslint file! 🚀

Bryant-Curtis: However if the eslint file is causing pains of sorts that I'm not aware of, I could see the value of not worrying too much about the 5 unsupported rules. Just wanted to raise the fact that those 5 rules might need to be commented out til supported unless they just happen to fail safely.

geoquant: good catch — you're right that those 5 aren't native in oxlint yet. they don't need to be though: the `jsx-a11y-js/` prefix runs the rules through oxlint's [JS plugin shim](https://oxc.rs/docs/guide/usage/linter/plugins#community-eslint-plugins), which loads the real `eslint-plugin-jsx-a11y` package at runtime. verified parity by running the same test file through both real ESLint and the oxlint JS plugin — identical hit counts across all 5 rules.

moved `aria-proptypes` and `no-static-element-interactions` to native `jsx-a11y/` as you suggested in bdc8854.

PR #234, opened 2026-03-16
feat: improve Input and Field components with dynamic sizing

leahwangcloudflare:
- Add size prop to Field component (xs, sm, base, lg)
- Dynamic gap spacing: lg uses gap-2, others use gap-1
- Dynamic label text size: xs=text-xs, sm=text-sm, base/lg=text-base
- Restructure label layout to show Optional text on right side
- Input passes size prop to Field for consistent sizing
- Update base input gap from gap-1.5 to gap-1


mattrothenberg (commented):
Thanks for the PR, Leah! The dynamic sizing feature is a solid improvement — having the Field automatically match the Input size makes the API much more intuitive.

A few suggestions:

---

**Gap sizing (field.tsx:31-32)**

Instead of the ternary chain, an object lookup is clearer and easier to extend if more sizes are added later:

```tsx
const gapBySize = {
  xs: "gap-1",
  sm: "gap-1", 
  base: "gap-1",
  lg: "gap-2",
} as const;

const gapClass = gapBySize[size];
```

---

**"Optional" text sizing (field.tsx:152)**

The label text scales with size (`text-xs`, `text-sm`, `text-base`), but "Optional" is always `text-sm`. When `size="xs"`, the label will be smaller than "Optional" which might look a bit off. Could use the same `labelTextSize` variable here for consistency.

---

**Layout change for "(optional)"**

Just flagging — this moves "(optional)" from inline (right after the label text) to right-aligned. That's a visual change for existing consumers. Totally fine if that's the intended design direction, just wanted to confirm it's intentional!

---

**CodeBlock examples cleared**

The `CodeBlock` examples array got emptied in the registry JSON — was that intentional or maybe an artifact from codegen?

---

Overall this is in good shape!

PR #273, opened 2026-03-20
feat(table): add Table.SortIcon and TanStack Table sortable/resizable demo

invisal:
## Summary

- Adds a new `Table.SortIcon` sub-component for use inside `Table.Head` to indicate sort direction.
- The `direction` prop accepts `false | "asc" | "desc"` — matching TanStack Table's `column.getIsSorted()` return type exactly, so consumers can pass the value directly without any conversion.
- Fixes `Table.ResizeHandle` to prevent click events from propagating to the column's sort handler after a drag.
- Adds a TanStack Table integration demo to the docs showing sortable columns and draggable column resizing working together.

## Usage

```tsx
// With TanStack Table — no conversion needed
const isSorted = header.column.getIsSorted(); // false | "asc" | "desc"
<Table.SortIcon direction={isSorted} />

// Standalone
<Table.SortIcon direction="asc" />
```


stritt:
thanks for putting this together! we're cleaning up the pr backlog and the sort icon and tanstack demo still look useful, but this needs a rebase that preserves the current table api, drops the old generated registry changes and fixes the lint error. are you still planning to work on it? if not we'll close it for now on july 28, but you're always welcome to reopen it when ready.

PR #282, opened 2026-03-22
Feat/timeseries a11y

jcworth:
## Summary

Kumo charts use `aria-enabled: true` which provides a default ECharts aria description, but this description is not reachable with keyboard navigation in order to read this out. This is because the chart container `<div>` has no tabindex or ARIA role.

Example from docs:

https://github.com/user-attachments/assets/caad0166-b4dc-48c8-8b34-2ce564b646b6

You'll notice that the mac screenreader will read it out when the mouse hovers during recording, but this isn't default behaviour and it is not possible to tab focus the chart.

Here's behaviour where we add tabindex and role="img"

https://github.com/user-attachments/assets/15cb148b-9dee-41fe-a66b-88532f110654

So we can now keyboard focus, and it reads the ECharts generated aria description. This PR also exposes an `ariaDescription` prop on Timeseries to override this with a custom readout.

## Changes
- Add tabIndex={0} and role="img" to the Chart container div when options.aria.enabled is truthy
- Add ariaDescription prop to TimeseriesChart for consumer-provided screen reader descriptions

### DIFF
--- a/.changeset/cyan-cobras-join.md
+++ b/.changeset/cyan-cobras-join.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": minor
+---
+
+Make Chart components keyboard focusable with tabindex. Timeseries adds `ariaDescription` prop 


packages/kumo/src/components/chart/TimeseriesChart.tsx
```diff
@@ -205,6 +216,7 @@ export function TimeseriesChart({
     return {
       aria: {
         enabled: true,
+        ...(ariaDescription && { label: { description: ariaDescription } }),
```
jcworth, line 219: there's a few other aria props available on echarts, maybe worth exposing the lot of them? https://echarts.apache.org/en/option.html#aria.label.description

Bryant-Curtis: Good question and good thinking!

I'd be curious to see how the behavior with the SR would change if we did.
Does adding them all make the readout more or less understandable?
Does it provide helpful detail, or is it more noise? Does the helpfulness/noise depend on what is added by consumers of this component, or the lack thereof?

And for clarity, answering these questions is not a show stopper for this MR.

packages/kumo/src/components/chart/TimeseriesChart.tsx
```diff
@@ -77,6 +77,16 @@ export interface TimeseriesChartProps {
    * that oscillates back and forth to indicate that data is being fetched.
    */
   loading?: boolean;
+  /**
+   * Accessible description for screen readers. When provided, it is passed to
+   * ECharts' `aria.label.description` and announced when the chart receives
+   * focus. Consumers are responsible for writing a meaningful description —
+   * see the W3C guidance on complex images for recommendations.
```
Bryant-Curtis, line 84: Spot on - thank you!

packages/kumo/src/components/chart/TimeseriesChart.tsx
jcworth: yeah to be honest i'm not entirely sure, this is something i'd like to follow up on after some experimentation 😄

PR #286, opened 2026-03-23
fix(sensitive-input): fix focus ring, defensive styles, and empty field typing

mattrothenberg:
## Summary

Fixes three papercuts in SensitiveInput:

1. **Focus ring not on the appropriate visual element** - When the inner input gains focus (revealed mode), the container now shows a browser-native focus outline via `focus-within`
2. **Eye icon inheriting global button styles** - Added defensive styles to prevent CSS pollution from consuming apps (e.g., Stratus's `cfBaseStyles`)
3. **Characters masked while typing into empty field** - When starting to type into an empty SensitiveInput, the component now switches to revealed mode so characters are visible immediately

### Focus ring (`sensitive-input.tsx`)

- Added `focus-within:outline focus-within:outline-2 focus-within:outline-[-webkit-focus-ring-color]` to container so the browser-native blue focus ring shows when the inner input is focused
- Added `outline-none` to inner input to suppress its own outline (container handles it)

### Defensive styles (`sensitive-input.tsx`)

- **Eye toggle button**: Added `bg-transparent border-none shadow-none p-0 m-0 h-auto min-h-0 inline-flex items-center justify-center`
- **Copy button**: Added `border-none shadow-none m-0 h-auto min-h-0`

### Empty field typing (`sensitive-input.tsx`)

- In `handleChange`, when the user types into an empty field (`mode === "empty"`), switch to `"revealed"` mode so the input renders as `type="text"` instead of `type="password"`

### inputVariants fix (`input.tsx`)

- Fixed `parentFocusIndicator` selector from `[&:has(:focus-within)]:ring-kumo-ring` to `focus-within:ring-kumo-ring` (the old selector was incorrect Tailwind)
- Also fixes InputGroup which uses `parentFocusIndicator` (though it already had a workaround)

## Testing

- Verified in Kumo docs dev server
- Designed for testing in Stratus kitchen sink page

## Related

- Follows defensive styles pattern from PR #214


packages/kumo/src/components/input/input.tsx
```diff
@@ -114,7 +114,7 @@ export function inputVariants({
     // Apply variant styles from KUMO_INPUT_VARIANTS
     KUMO_INPUT_VARIANTS.variant[variant].classes,
     // Focus state handling
-    parentFocusIndicator && "[&:has(:focus-within)]:ring-kumo-ring",
```
mattrothenberg, line 117: I think this was a bug.

PR #291, opened 2026-03-24
LayerCard secondary layout adjustments

Brayden:
### Problem

When composing LayerCard in a Secondary → Primary → Secondary order, the second Secondary rendered on top of the Primary, visually hiding the primarys bottom border. Also when a Secondary had a background color it would not be visible going underneath the Primary element so the corners would lack color.

This happened because LayerCardSecondary uses -my-2 (negative vertical margin) to bleed into adjacent elements, but without an explicit stacking context, later DOM elements paint over earlier ones by default.

Another issue was the X padding on the Secondary element did not match the X padding on the Primary. This change gives them the same padding treatment.

### Solution

Added relative to LayerCardPrimary so it always establishes a stacking context above adjacent Secondary elements, regardless of DOM order.


packages/kumo/src/components/layer-card/layer-card.tsx
```diff
@@ -55,7 +55,7 @@ function LayerCardSecondary({ children, className }: LayerCardProps) {
   return (
     <div
       className={cn(
-        "flex items-center gap-2 p-2 text-base font-medium text-kumo-strong bg-kumo-recessed",
+        "flex items-center gap-2 p-4 text-base font-medium text-kumo-strong -my-2 bg-kumo-recessed",
```
mattrothenberg, line 58: The jump from `p-2` to `p-4` will probably lead to some visual changes downstream. Any concerns there?

Brayden: Looking at a few places where it's been used (internally) at least it seems like it will help more than hurt aligning the text between sections. Keen to believe a positive impact overall, at least that's the optimistic hope!

mattrothenberg: Wonderful, that's a great outcome.

PR #295, opened 2026-03-24
feat(tabs): animate indicator with Framer Motion spring transform

nandanmen:
This PR updates the tab component to animate its transition using `motion/react` and a `translate` transform. This offers a few improvements:

- transforms are run on the compositor thread so it should be more performant
- `motion/react` automatically optimizes its animation implementation to use WAAPI whenever possible
- spring animation makes the transition feel a bit more natural


packages/kumo/src/components/tabs/tabs.tsx
```diff
@@ -99,6 +100,8 @@ export type TabsProps = KumoTabsVariantsProps & {
   indicatorClassName?: string;
 };
 
+const TabIndicator = motion(TabsPrimitive.Indicator);
```
mattrothenberg: unused? could we actually do this instead of the `render` prop entirely?

nandanmen: oh yeah good point, should remove. I think the render approach is better, it's easier to reason about

PR #310, opened 2026-03-26
Feat/timeseries focus series

jcworth:
## Summary
When a `TimeseriesChart` has multiple series, hovering over a line gives no visual feedback about which series you're inspecting — all lines remain at equal opacity. 

ECharts supports `emphasis: { focus: "series" }` on series options, which automatically dims all other series when hovering over a line. This feels like a sensible default for any multi-series chart, so we enable it on all series entries (including incomplete/dashed segments, and bar charts).

## Changes
- Add `emphasis: { focus: "series" }` to main series and incomplete (dashed) series configs in `TimeseriesChart`
- Hovering a line now highlights it and fades all other series

## 

Old behaviour:
oldhover

New behaviour:
charthover

### DIFF
--- a/.changeset/eager-pens-work.md
+++ b/.changeset/eager-pens-work.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": minor
+---
+
+Make timeseries' highlighted on hovering a data point


invisal:
> I like this! @invisal WDYT?

Look good to me. ^^

PR #330, opened 2026-03-30
feat(chart): chart legend item series highlight

jcworth:
Fixes N/A.

Building on https://github.com/cloudflare/kumo/pull/310 (series emphasis on line hover) adds the event handling needed for legend-to-chart highlight interactions.

`ChartLegend` items and `TimeseriesChart` are fully decoupled — there's no way to hover a legend item and have the corresponding series highlight on the chart even though Timeseries supports highlighting the series directly.

Adds `onPointerEnter`/`onPointerLeave` props to legend items and wraps `TimeseriesChart` in `forwardRef` so consumers can get a ref to the ECharts instance and call `dispatchAction` for highlight.

reference: https://apache.github.io/echarts-handbook/en/concepts/event/

Changes:

- Add `onPointerEnter` and `onPointerLeave` props to `ChartLegend.LargeItem` and `ChartLegend.SmallItem`
- Wrap `TimeseriesChart` in `forwardRef` to expose the ECharts instance (uses merged callback ref pattern from `SensitiveInput`)
- Add `LegendHighlightDemo` to the docs site showing the full pattern
- Add "Legend Highlight" section to the timeseries docs page

Usage:

```tsx
const chartRef = useRef<echarts.ECharts>(null);
const [hoveredSeries, setHoveredSeries] = useState<string | null>(null);
<ChartLegend.LargeItem
  name="P99"
  inactive={hoveredSeries !== null && hoveredSeries !== "P99"}
  onPointerEnter={() => {
    setHoveredSeries("P99");
    chartRef.current?.dispatchAction({ type: "highlight", seriesName: "P99" });
  }}
  onPointerLeave={() => {
    setHoveredSeries(null);
    chartRef.current?.dispatchAction({ type: "downplay", seriesName: "P99" });
  }}
/>
<TimeseriesChart ref={chartRef} ... />
```

Demo:

legend-highlight

---

- Reviews
- Tests
  - [ ] Tests included/updated
  - [x] Additional testing not necessary because: this is a small implementation change


mattrothenberg (approved):
LGTM. @invisal can I get your gut check?

packages/kumo/src/components/chart/Legend.tsx
```diff
@@ -12,16 +13,32 @@ interface LegendItemProps {
   unit?: string;
   /** When `true`, renders the item at 50% opacity to indicate a deselected state */
   inactive?: boolean;
+  /** Fired when a pointer enters the legend item — useful for highlighting the corresponding chart series */
+  onPointerEnter?: PointerEventHandler<HTMLDivElement>;
+  /** Fired when a pointer leaves the legend item — useful for resetting chart series emphasis */
+  onPointerLeave?: PointerEventHandler<HTMLDivElement>;
```
invisal, line 29: Should we also expose `onClick` and `className`? I can see a use case where people want to support **legend filtering** by clicking on the legend. They will likely want to override `cursor-pointer` to the className.

jcworth: hadn't thought about that, good shout. I'd done an implementation where i'm rendering filter buttons alongside the legend but seems like a valid use case. will update

jcworth: @invisal re: classname,  might it just be worth exposing a prop like `pointer`? if we add classname it kind of opens door to users overriding a variety of things on something that is fairly opinionated (unless we see this as desirable / a feature.?)

invisal: @jcworth 
There is always value in exposing the className. The majority of our components have an exposed className. It acts as a last resort and does not block our developers' work while they wait for us to cover their use cases.

If we want to be opinionated, we can add cursor-pointer when onClick is specified; this saves us one prop.

jcworth:
> Hey @jcworth -- lint is failing with 4 a11y errors in `Legend.tsx`. The clickable `<div>` elements at lines 43 and 101 need a `role="button"` attribute and an `onKeyDown` handler (to handle Enter/Space). Should be a quick fix, then this is good to merge.

would it be preferable to conditionally render it as a button (dependent on onClick being passed) rather than applying role to a div? else might have to override the linter `prefer-tag-over-role`. it could _always_ be a button but would need to tabIndex -1 it when theres no onclick

@invisal what do you reckon?

invisal:
> > Hey @jcworth -- lint is failing with 4 a11y errors in `Legend.tsx`. The clickable `<div>` elements at lines 43 and 101 need a `role="button"` attribute and an `onKeyDown` handler (to handle Enter/Space). Should be a quick fix, then this is good to merge.
> 
> would it be preferable to conditionally render it as a button (dependent on onClick being passed) rather than applying role to a div? else might have to override the linter `prefer-tag-over-role`. it could _always_ be a button but would need to tabIndex -1 it when theres no onclick
> 
> @invisal what do you reckon?

Yes. We can just add `role='button'` at the moment

invisal:
It looks good to me. @mattrothenberg @jcworth

PR #336, opened 2026-03-31
feat(SkeletonLine): expose height prop via --skeleton-height CSS variable

geoquant:
## Problem

The `.skeleton-line` CSS class in `kumo-binding.css` uses `var(--skeleton-height, 0.5rem)`, but the `SkeletonLine` component never sets this CSS variable. This means consumers have no way to control skeleton loader height through the component API — it's always locked to the `0.5rem` fallback.

## Solution

Add an optional `height` prop to `SkeletonLine` that maps to the `--skeleton-height` CSS custom property, following the same pattern already used for `--skeleton-width`.

```tsx
// Now consumers can set different heights:
<SkeletonLine height="1rem" />
<SkeletonLine height="24px" />
<SkeletonLine />  // still defaults to 0.5rem
```

## Changes

- **`packages/kumo/src/components/loader/skeleton-line.tsx`**: Added `height?: string` prop, conditionally spreads `--skeleton-height` into the inline style object
- Non-breaking — omitting `height` preserves the existing `0.5rem` default via CSS fallback

## Checklist

- [x] Typecheck passes
- [x] Lint passes
- [x] Changeset included

### DIFF
--- a/.changeset/badge-color-variants.md
+++ b/.changeset/badge-color-variants.md
@@ -1,0 +1,12 @@
+---
+"@cloudflare/kumo": minor
+---
+
+Badge: add color-based variants and deprecate semantic variants
+
+- Add color variants: `red`, `orange`, `yellow`, `green`, `teal`, `blue`, `neutral`, `inverted`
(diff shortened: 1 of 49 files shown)


bradgarropy:
Looks great to me!

PR #350, opened 2026-04-01
fix(select): restore touch-open popup scrolling

singkia:
## What

- Restore the missing scroll-container semantics for touch-open `Select` menus.
- Keep the existing popup visuals unchanged while constraining the popup with Base UI's `--available-height` viewport calculation.
- Add a regression test that fails if the popup falls back to the old `max-h-full` behavior.
- Add a patch changeset for `@cloudflare/kumo` so the branch passes the repo's changeset validation.

## Why

This bug only shows up on the touch-open path, not the usual desktop mouse-open path.

When Base UI opens `Select` from touch input, it disables `alignItemWithTriggerActive`. In that branch, Base UI no longer injects the functional scroll-container styles that normally keep tall menus constrained and internally scrollable.

Kumo's `Select` popup still hard-codes its own overflow behavior, so long option lists are treated as one oversized popup instead of a viewport-constrained container with internal scrolling.

That means the bug only becomes visible when both of these conditions are true:

- the menu is opened through the touch path
- the option list is tall enough to require scrolling

Short menus do not visibly break because they never need to scroll. Desktop menus usually avoid the issue because they use a different open path.

This invalid state showed up in a downstream consumer as unscrollable mobile menus for long option lists. The fix restores the missing popup scroll-container semantics in `Select` itself and now uses Base UI's `--available-height` variable so the popup respects the actual viewport space calculated by `Positioner`.

## Before/After

**Before**

https://github.com/user-attachments/assets/732455f0-fe8d-4236-9cd1-0440caeab2c2

**After**

https://github.com/user-attachments/assets/75aa890d-2988-4bcf-af9b-d752b5392f05

## QA Steps

1. Open a consumer app with a long `Select` menu on mobile or touch input.
2. Trigger a tall menu such as `Region` or `Device`.
3. Verify the popup stays inside the viewport and scrolls internally.
4. Verify desktop mouse-open behavior is visually unchanged.

## Validation Checklist

- [x] automated review not possible because: `/bonk review` was triggered, but Bonk failed with `UnknownError: ProviderInitError` instead of producing an automated review.
- [x] Tests included/updated

## Test Results

```bash
pnpm --filter @cloudflare/kumo exec vitest run --project=unit src/components/select/select.test.tsx
pnpm --filter @cloudflare/kumo lint
pnpm --filter @cloudflare/kumo typecheck
pnpm tsx ci/scripts/validate-kumo-changeset.ts
```

The updated branch was also verified in a local `jikan.life` consumer worktree with the page-level CSS workaround removed, confirming the `Region` and `Device` menus still scroll correctly when driven only by the Kumo package fix.

---

## AI disclosure

This PR was prepared with OpenAI Codex.

Prompts given to the agent:
- Inspect the `jikan.life` fix and determine whether the same root cause applies to Kumo `Select`.
- Verify the fix in isolated worktrees without modifying the original repos.
- Restore the minimal `Select` fix in `kumo`, add a regression test, run the required local checks, and draft a normalized PR description using the Gumroad CONTRIBUTING structure as a reference.

### DIFF
--- a/.changeset/fix-codeblock-unknown-lang.md
+++ b/.changeset/fix-codeblock-unknown-lang.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Fix CodeBlock crash when an unsupported `lang` value is passed at runtime. The `codeVariants()` function now uses optional chaining with a nullish coalescing fallback to the default language, matching the defensive pattern already used by `switchVariants()` and `badgeVariants()`.
--- a/.changeset/tooltip-delay-props.md
+++ b/.changeset/tooltip-delay-props.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": minor
+---
+
+Expose `delay` and `closeDelay` props on Tooltip component to control open and close timing. These props are forwarded to the underlying Base UI Trigger component.
--- a/.github/workflows/bonk-check.yml
+++ b/.github/workflows/bonk-check.yml
@@ -11,7 +11,6 @@ on:
 permissions:
   contents: read
   pull-requests: read
-  statuses: write
 
 jobs:
   bonk-check:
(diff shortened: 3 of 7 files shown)


singkia:
Self-review:

- Root cause: on the touch-open path, Base UI stops applying the scroll-container styles that normally constrain tall Select menus, while Kumo still forces `overflow-hidden` on the popup.
- Why this only shows up sometimes: the bug only becomes visible when the menu is opened through touch input and the option list is tall enough to require scrolling. Short menus do not visibly break, and desktop mouse-open usually takes a different path.
- Fix scope: this change restores the missing popup scroll semantics (`relative`, `max-h-full`, `overflow-x-hidden`, `overflow-y-auto`) without changing the popup's visual styling.
- Regression coverage: added a unit test that fails if the popup falls back to the old `overflow-hidden`-only behavior.
- Risk checked: the main risk is unintended desktop behavior drift, so the change was kept minimal and scoped to restoring the missing scroll-container behavior rather than changing layout logic higher up the stack.

mattrothenberg (changes requested):
Suggestion: Use Base UI's `--available-height` CSS variable

The current fix uses `max-h-full`, but this isn't quite the same as what Base UI provides. Base UI's `Positioner` calculates the actual available viewport space and exposes it via `--available-height`.

**Current:**
```tsx
"relative max-h-full overflow-x-hidden overflow-y-auto bg-kumo-base text-kumo-default"
```

**Suggested:**
```tsx
"relative max-h-[var(--available-height)] overflow-x-hidden overflow-y-auto bg-kumo-base text-kumo-default"
```

**Why this matters:**
- `max-h-full` = 100% of parent height (may expand to fit content if parent isn't constrained)
- `--available-height` = Actual calculated viewport space from Base UI's collision detection

This is especially important for mobile edge cases (small viewports, keyboard open, etc.) where the Positioner calculates less available space than "full" would imply.

---

**Docs suggestion:**

It would be great to add a documentation example that exercises this condition — a `Select` with a long option list (30+ items) rendered in a constrained or mobile viewport. This would give us visual confidence that the scrolling behavior works as expected across both desktop and touch paths, and prevent future regressions.

singkia:
@mattrothenberg Thanks for flagging this. We were able to reproduce the rubberbanding issue from your screenshot with a long `Select` menu.

I pushed a minimal follow-up fix that adds `overscroll-y-none` to the popup scroll container. The goal is to stop the edge overscroll behavior that was letting the underlying page show through when the menu was pulled past its scroll bounds. After that change, we’re no longer seeing the bleed on our side.

When you have a moment, could you re-test on the original device / interaction path and confirm whether this resolves it?

packages/kumo/src/components/select/select.tsx
```diff
@@ -1,3 +1,9 @@
+/**
+ * [INPUT]: 依赖 @base-ui/react/select 的弹出层与选项能力，依赖 button/field/loader/portal-provider 的 Kumo 样式与容器能力
+ * [OUTPUT]: 对外提供 Select 组件、Select.Option 子组件、KUMO_SELECT_* 元数据与样式工具
+ * [POS]: components/select 的核心实现，负责触发器、弹层、选项渲染与 Field 集成
+ * [PROTOCOL]: 变更时更新此头部，然后检查 CLAUDE.md
+ */
```
mattrothenberg: please drop

singkia: Dropped, thanks.

PR #356, opened 2026-04-02
fix(ci): stop bonk-check from failing on status writes

geoquant:
Fixes failing Bonk Check job: https://github.com/cloudflare/kumo/actions/runs/23924593430/job/69779025414

The Bonk Check workflow was failing in `Post commit status` with `Resource not accessible by integration (HTTP 403)` when it tried to create a commit status from PR/review contexts.

This change removes the manual commit-status API call and lets the workflow job itself be the required pass/fail signal instead. It also fixes a false-negative on `pull_request_review`: the old script only scanned issue comments, so `/bonk` in a review body did not satisfy the check.

Specifically, this PR:
- removes `statuses: write` from the workflow permissions
- drops the `gh api .../statuses/...` step that was causing the 403
- fails the workflow normally when `/bonk` has not been called by a collaborator
- recognizes `/bonk` in review bodies as well as PR comments
- uses more robust JSON/base64 parsing for comment bodies instead of `sed` splitting

---

- Reviews
  - [x] automated review not possible because: this PR fixes the Bonk check workflow itself, and the failing check is the thing being changed
- Tests
  - [ ] Tests included/updated
  - [x] Additional testing not necessary because: the change is isolated to GitHub Actions workflow logic and directly removes the failing 403 status-write path confirmed in the job logs

### DIFF
--- a/.github/workflows/bonk-check.yml
+++ b/.github/workflows/bonk-check.yml
@@ -11,7 +11,6 @@ on:
 permissions:
   contents: read
   pull-requests: read
-  statuses: write
 
 jobs:
   bonk-check:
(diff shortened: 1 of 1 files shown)


geoquant:
Closing in favor of a single self-contained follow-up PR that replaces the manual /bonk gate with automatic Bonk reviews on PR open/update.

PR #390, opened 2026-04-10
fix(flow): disable panning/dragging when canvas=false

nandanmen:
Fixes a bug where text inside a flow node is not selectable even when `canvas` is set to `false` in the parent component (therefore disabling all dragging features).

## Changes

- When `canvas={false}`, pan event handlers (`onPanStart`, `onPan`, `onPanEnd`) are no longer attached to the motion wrapper, so dragging/panning is fully inert
- The `grab`/`grabbing` cursor and `userSelect: none` body mutations are suppressed — `canvas && canPan && !isPanning` guards the cursor style, and the handlers that set body styles are never called
- Adds `FlowWithCodeBlockDemo` to the `/tests/flow` test page: a `canvas={false}` flow where one node embeds a `CodeHighlighted` block, verifying syntax highlighting works inside a flow node and that text selection is not blocked


mattrothenberg (approved):

stritt:
thanks for this! the underlying canvas=false fix still looks useful, but the branch now conflicts with the current flow implementation. this may be easiest to port onto current main with the tests, but feel free to rebase it if you have time and we can get it reviewed again.

PR #397, opened 2026-04-13
feat(link): introduce `useClientRouting` hook to for router integration

1000hz:
This PR introduces a new `useClientRouting({onNavigate, onPrefetch})` hook to allow Kumo to integrate with client-side routers by listening for `kumo:navigate` and `kumo:prefetch` custom events dispatched from `Link` (on `click` and `mouseenter`/`focus` respectively).

```tsx
import { useNavigate } from "react-router";
import { useClientRouting } from "@cloudflare/kumo";

function App() {
  const navigate = useNavigate();

  useClientRouting({
    onNavigate: (href, event) => {
      if (URL.canParse(href)) {
        event.preventDefault(); // cancel the native browser navigation!

        const { pathname, search, hash } = URL.parse(href);
        const anchor = event.detail.anchor;

        navigate(
          { pathname, search, hash },
          { replace: anchor.dataset.method === "replace" }
        );
      }
    }
  });
}
```

The custom events are dispatched only for links to internal hrefs (i.e. same origin, doesn't have a non-`_self` target, no `download` attribute), and only when they are activated normally (i.e. no modifier keys were held, no middle-click / right-click activation).

---

- Reviews
- Tests
  - [x] Tests included/updated

### DIFF
--- a/.changeset/calm-rivers-share.md
+++ b/.changeset/calm-rivers-share.md
@@ -1,0 +1,8 @@
+---
+"@cloudflare/kumo": minor
+---
+
+feat(link): add router-agnostic client routing events
+
+- Emit `kumo:navigate` and `kumo:prefetch` from eligible internal links
+- Add `useClientRouting()` to integrate Link events with application routers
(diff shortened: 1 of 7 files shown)


1000hz:
after internal discussion, we decided it's best to keep this sort of logic strictly in userland (i.e. via a custom `Link` given to `LinkProvider`) rather than within Kumo itself.

PR #411, opened 2026-04-14
Add copy to clipboard to all code blocks

pedromenezes1:
Adds a copy-to-clipboard button to every code block in the docs site.

Users frequently need to copy code snippets from the docs (install commands, usage examples, API references) but had no way to do so without manually selecting text. The button makes this a single click.

https://github.com/user-attachments/assets/159a91ef-9bdc-4fce-8aea-8efcda5b91e3

---

- Reviews
  - [x] automated review not possible because: docs-only change, no modifications to the published `@cloudflare/kumo` component library
- Tests
  - [ ] Tests included/updated
  - [x] Additional testing not necessary because: docs-only content changes

### DIFF
--- a/packages/kumo-docs-astro/src/components/docs/CodeBlock.astro
+++ b/packages/kumo-docs-astro/src/components/docs/CodeBlock.astro
@@ -9,9 +9,7 @@ interface Props {
 const { code, lang = "tsx" } = Astro.props;
 ---
 
-<div
-  class="not-prose code-block"
->
+<div class="not-prose code-block relative">
   <Code 
     code={code} 
     lang={lang} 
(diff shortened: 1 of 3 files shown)


packages/kumo-docs-astro/src/components/docs/CodeBlock.astro
```diff
@@ -21,27 +19,22 @@ const { code, lang = "tsx" } = Astro.props;
     }}
     defaultColor={false}
   />
+  <button
+    type="button"
+    class="code-block-copy-btn"
```
pedromenezes1, line 24: It copies the style of our `<Button variant="ghost" size="sm" />` button.

packages/kumo-docs-astro/src/components/docs/CodeBlock.astro
```diff
@@ -21,27 +19,22 @@ const { code, lang = "tsx" } = Astro.props;
...
+    aria-label="Copy code to clipboard"
+  >
+    <span class="code-block-copy-btn__icons">
+      <span class="code-block-copy-btn__icon code-block-copy-btn__icon--check" aria-hidden="true">
+        <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" viewBox="0 0 256 256">
```
pedromenezes1, line 29: It's Phosphor icons, just couldn't use the React library here.

mattrothenberg (approved):

PR #413, opened 2026-04-15
fix(select): make renderValue and placeholder work together

mattrothenberg:
## Summary

This PR fixes two issues with the Select component:

1. **TypeScript inference bug** - Under `strictNullChecks`, using `value={objectOrNull}` would cause `T` to be inferred as `never`, making callback props like `renderValue` and `onValueChange` unusable.

2. **`placeholder` ignored when using `renderValue`** - Base UI's design means the `placeholder` prop is completely ignored when you provide a children-as-function (which `renderValue` maps to). Users had to handle the null case inside `renderValue`, defeating the purpose of `placeholder`.

## The Fix

Kumo now wraps the `renderValue` function internally to:
- Show `placeholder` (with proper styling) when value is null
- Only call user's `renderValue` with non-null values
- If no `placeholder` is provided and value is null, return `null` (show nothing)

```tsx
// Recommended pattern
<Select
  placeholder="Select a user..."
  value={user}
  renderValue={(user) => user.name}  // Only called when user is not null
/>
```

### `packages/kumo/src/components/select/select.tsx`
- Wrap `renderValue` to handle null internally, showing placeholder with proper styling
- When no placeholder provided and value is null, return null (no empty span)
- Simplify valueChildrenFn logic

### `packages/kumo-docs-astro/`
- Update demos to use `placeholder` + `renderValue` pattern
- Add comprehensive stress test page at `/tests/select` with 12 scenarios
- Update documentation

### `packages/kumo/tests/`
- Add TypeScript type tests with `strictNullChecks` and `strictFunctionTypes` enabled
- Verify type inference works correctly for all use cases (including without explicit generic)

## Stress Test Page

Visit `/tests/select` to see all API permutations:
1. strictNullChecks with object values
2. renderValue + placeholder interaction
3. items array prop
4. items object map with disabled options
5. Multiple selection
6. Controlled vs uncontrolled
7. Groups and separators
8. Field integration (label, error, description)
9. Size variants
10. Loading state
11. Disabled options
12. Type inference without explicit generic

## Testing

- [x] TypeScript compiles in kumo
- [x] TypeScript compiles with stratus-equivalent strict settings
- [x] Type tests pass without explicit `<Select<T>>` generic syntax
- [x] Verified `placeholder` shows when value is null
- [x] Verified `renderValue` is only called with non-null values
- [x] Verified no empty span when placeholder is not provided
- [x] All 12 stress test scenarios work correctly

---

- Reviews
- [x] automated review not possible because: this is a behavioral fix that requires human judgment on the API design
- Tests
- [x] Tests included/updated
- [x] Additional testing not necessary because: type tests and stress test page added


mattrothenberg:
/bonk I need a thorough review of this one.

geoquant (approved):

mattrothenberg:
/bigbonk OK I made the changes you requested

PR #415, opened 2026-04-15
fix(Checkbox): preserve native event APIs in onChange compat layer

### LINKED ISSUE #409: Checkbox: Object.assign on read-only Event.target crashes onCheckedChange
## Description

The `Checkbox` component crashes when clicked due to `Object.assign(event, { target: ... })` in the `onCheckedChange` handler. `Event.target` is a read-only getter property and cannot be set via `Object.assign`.

## Error

```
Uncaught TypeError: Cannot set property target of #<Event> which has only a getter
    at Object.assign (<anonymous>)
    at Object.onCheckedChange (@cloudflare_kumo.js:4043:32)
```

## Steps to Reproduce

1. Render a `<Checkbox>` component
2. Click on it
3. The `onCheckedChange` handler throws

## Environment

- `@cloudflare/kumo` v1.18.0
- React 19.2.4
- Chrome (latest)

## Expected Behavior

Checkbox should toggle without errors.

## Actual Behavior

Throws `TypeError: Cannot set property target of #<Event> which has only a getter` on every click. The checkbox state does not change.

singkia:
## What

Fix the deprecated `onChange` compatibility path on `Checkbox` so it no longer crashes on read-only `Event.target` and still preserves native event APIs.

## Why

`Object.assign(eventDetails.event, { target: { checked: newChecked } })` tries to write to `Event.target`, which is a read-only getter on native DOM events and crashes in modern browsers.

The first fix using `Object.create(eventDetails.event, { target: ... })` avoided that write, but it also stopped passing a real native event object through the deprecated `onChange` path. Brand-checked APIs like `preventDefault()`, `defaultPrevented`, and `timeStamp` could then break.

This version proxies the original event and only overrides `target.checked` and `currentTarget.checked`, while binding the rest of the event API back to the native event object.

## Test Results

Added a focused unit test for the deprecated `onChange` compatibility path that verifies:

- `target.checked` and `currentTarget.checked` reflect the new value
- native event APIs like `preventDefault()`, `defaultPrevented`, and `timeStamp` still work

Commands run:

- `pnpm --filter @cloudflare/kumo exec vitest run src/components/checkbox/checkbox.test.tsx`
- `pnpm --filter @cloudflare/kumo exec tsc --noEmit`
- `pnpm --filter @cloudflare/kumo test` (still hits the pre-existing `tests/imports/export-path-validation.test.ts` failures for `table-of-contents` dist artifacts)

Fixes #409

---

AI disclosure: Claude Opus 4.6 (via Amp) authored the original patch; GPT 5.4 (via Codex) updated the fix after review to preserve native event semantics and added regression coverage.

- Reviews
- [x] automated review not possible because: external contributor without access to bonk
- Tests
- [x] Tests included/updated

### DIFF
--- a/.changeset/fix-checkbox-onchange-target.md
+++ b/.changeset/fix-checkbox-onchange-target.md
@@ -2,7 +2,7 @@
 "@cloudflare/kumo": patch
 ---
 
-fix(Checkbox): use Object.create instead of Object.assign to avoid crashing on read-only Event.target
+fix(Checkbox): use Proxy instead of Object.assign to avoid crashing on read-only Event.target
 
 The deprecated `onChange` handler used `Object.assign(event, { target: ... })` which throws
 `TypeError: Cannot set property target of #<Event> which has only a getter` because `Event.target`
--- a/.changeset/fix-checkbox-onchange-target.md
+++ b/.changeset/fix-checkbox-onchange-target.md
@@ -1,0 +1,12 @@
+---
+"@cloudflare/kumo": patch
+---
+
+fix(Checkbox): use Object.create instead of Object.assign to avoid crashing on read-only Event.target
+
+The deprecated `onChange` handler used `Object.assign(event, { target: ... })` which throws
+`TypeError: Cannot set property target of #<Event> which has only a getter` because `Event.target`
+is a read-only getter property. Replaced with `Object.create` to create a new object that shadows
+the prototype getter with an own `target` property.
+
+Fixes #409
--- a/.opencode/package-lock.json
+++ b/.opencode/package-lock.json
@@ -5,21 +5,100 @@
   "packages": {
     "": {
       "dependencies": {
-        "@opencode-ai/plugin": "1.4.0"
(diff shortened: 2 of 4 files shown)


mattrothenberg:
/bigbonk will this crash in older browsers? we have to support a LOT of browsers

mattrothenberg:
@singkia If I were to tell you that we're planning on cutting a major release of Kumo that includes breaking changes (e.g., getting rid of `onChange` altogether) would that change your approach? Is this change worth it in light of that? I appreciate you taking a look at this!

singkia:
> @singkia If I were to tell you that we're planning on cutting a major release of Kumo that includes breaking changes (e.g., getting rid of `onChange` altogether) would that change your approach? Is this change worth it in light of that? I appreciate you taking a look at this!

Thanks for the heads-up, @mattrothenberg. That makes sense! Saw you've already got #416 removing `onChange` entirely — that's the cleaner solution. Happy to close this one.

mattrothenberg:
@singkia Thank you. I appreciate you looking into this and fixing it.

PR #416, opened 2026-04-15
feat(checkbox)!: align onCheckedChange with Base UI, remove deprecated props

mattrothenberg:
## Summary

This PR aligns Checkbox's `onCheckedChange` callback with Base UI's native signature and removes deprecated props that have been superseded.

## Background

Base UI's Checkbox provides `onCheckedChange` with two arguments:
```tsx
onCheckedChange?: (checked: boolean, eventDetails: CheckboxRootChangeEventDetails) => void
```

Kumo was wrapping this and only exposing the first argument (`checked: boolean`), while maintaining three deprecated alternative props (`onChange`, `onValueChange`, `onClick`) for backwards compatibility. This added complexity and prevented consumers from accessing the underlying event when needed.

### `onCheckedChange` now exposes event details

The callback signature now matches Base UI:

```tsx
// Before
onCheckedChange={(checked) => console.log(checked)}

// After - eventDetails available as optional second argument
onCheckedChange={(checked, eventDetails) => {
  console.log(checked);
  console.log(eventDetails.event); // native event if needed
}}
```

Existing code that only uses the first argument continues to work unchanged.

### Removed deprecated props

| Removed Prop | Replacement |
|--------------|-------------|
| `onChange` | `onCheckedChange` |
| `onValueChange` | `onCheckedChange` |
| `onClick` | Standard React event handling via spread props |

### Simplified internals

- Removed `handleCheckedChange` wrapper functions in both `CheckboxBase` and `CheckboxItem`
- Now passes `onCheckedChange` directly to Base UI
- Type is derived directly from Base UI: `BaseCheckbox.Root.Props["onCheckedChange"]`

### New export

Added `CheckboxChangeEventDetails` type export for consumers who need to type the event details:

```tsx
import { Checkbox, type CheckboxChangeEventDetails } from '@cloudflare/kumo';

const handleChange = (checked: boolean, details: CheckboxChangeEventDetails) => {
  // ...
};
```

## What's NOT changing

- **`Checkbox.Group`'s `onValueChange` prop** - This is a different API that returns `string[]` (the array of selected values), not a boolean. It remains unchanged.
- **`Table.CheckCell` and `Table.CheckHead`** - These have their own `onValueChange` prop that wraps Checkbox internally. They continue to provide a simpler `(checked: boolean) => void` API, which makes sense for table selection use cases.

## Migration

```tsx
// Before (deprecated)
<Checkbox onChange={(e) => console.log(e.target.checked)} />
<Checkbox onValueChange={(checked) => setChecked(checked)} />

// After
<Checkbox onCheckedChange={(checked) => setChecked(checked)} />

// If you need the event
<Checkbox onCheckedChange={(checked, { event }) => {
  console.log(event);
}} />
```

## Breaking Change Justification

1. **The deprecated props have been marked deprecated** - consumers have had time to migrate
2. **Simpler API surface** - one callback instead of three overlapping ones
3. **Better Base UI alignment** - reduces maintenance burden and gives consumers access to the full API
4. **Type safety** - the `onChange` prop was using `as never` cast internally, which was a code smell

---

- Reviews
- [x] automated review not possible because: this is a breaking API change requiring human judgment
- Tests
- [x] Tests included/updated
- [x] Additional testing not necessary because: existing checkbox demos use onCheckedChange and continue to work; API table auto-generates from types

### DIFF
--- a/.changeset/checkbox-breaking-onCheckedChange.md
+++ b/.changeset/checkbox-breaking-onCheckedChange.md
@@ -1,0 +1,37 @@
+---
+"@cloudflare/kumo": major
+---
+
+**BREAKING:** Checkbox `onCheckedChange` now receives event details as second argument
+
+The `onCheckedChange` callback signature now matches Base UI, providing access to the underlying event:
+
+```tsx
+// Before
+onCheckedChange={(checked) => console.log(checked)}
+
+// After (event details available as optional second arg)
+onCheckedChange={(checked, eventDetails) => {
+  console.log(checked);
+  console.log(eventDetails.event); // native event
+}}
+```
+
+**Removed deprecated props:**
+
+- `onChange` - use `onCheckedChange` instead
+- `onValueChange` on individual checkboxes - use `onCheckedChange` instead
+- `onClick` - was redundant, use standard React event handling via spread props
+
+**Migration:**
+
+```tsx
+// Before (deprecated)
+<Checkbox onChange={(e) => console.log(e.target.checked)} />
+<Checkbox onValueChange={(checked) => setChecked(checked)} />
+
+// After
+<Checkbox onCheckedChange={(checked) => setChecked(checked)} />
+```
+
+Note: `Checkbox.Group`'s `onValueChange` prop is unchanged - it still accepts `(values: string[]) => void`.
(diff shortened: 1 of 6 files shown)


mattrothenberg:
/bigbonk updated. can you re-review and approve now?

geoquant (approved):

PR #426, opened 2026-04-16
feat(tokens): chart color fixes

najlaskr:
## Summary

Adds a dedicated chart colors documentation page and refines chart color demos + palette ordering for clearer, more consistent color guidance.

## What changed

- Added new chart colors docs page
- Updated chart color demos
- Updated chart demo typing/behavior cleanup
- Updated charts index page to include the new colors docs entry
- Updated docs sidebar navigation for charts/colors
- Updated chart categorical palette ordering in library code

## Why

- Improves chart color documentation clarity and discoverability.
- Aligns categorical palette ordering across docs and implementation.
- Keeps chart demo behavior and docs examples consistent with current chart APIs.

## Validation

- `pnpm lint` ✅
- `pnpm typecheck` ✅

Reviews

Tests
- [ ] Tests included/updated
- [ ] Automated tests not possible - manual testing has been completed as follows: N/A
- [x] Additional testing not necessary because: changes are docs/demo and palette ordering updates validated via lint and typecheck


geoquant:
looks good! can we add it so that it is visible with the command palette? 
Screenshot 2026-04-17 at 7 00 46 AM

geoquant:
also, i thought we needed a palette with 16 color steps?

geoquant:
also fwiw i think our "semantic" colors are either "diverging" or "categorical"
Screenshot 2026-04-17 at 7 15 29 AM


so we have 1 sequential palette and 2 categorical/qualitative palettes

but i'm not a purist, i just want us to build out an api that is flexible enough to support many color palettes under a type. like we may want a range of blues, under sequential.

check out this api:
https://observablehq.com/@d3/color-schemes

geoquant (changes requested):
left a few comments!

geoquant:
this api locks us into a palette, i'd prefer something like this:

ChartPalette.sequential('blues', 0, isDarkMode)

this would also address that the two proposed "categorical" and "semantic" colors are both qualitative palettes:

ChartPalette.qualitative('semantic', 0, isDarkMode)
ChartPalette.qualitative('categorical', 0, isDarkMode)

najlaskr:
made a new commit with these updates:

- updated docs with why we limit to only 5 colours and if we want more we can use modulo
- updated API to be clearer with colour use cases. I dont think we'd want to support new colour palettes for now but I've made the API flexible enough if we do want to add more in the future.

geoquant (approved):

PR #445, opened 2026-04-23
Docs: Fix copy code button appearing above header on scroll

pedromenezes1:
The copy button inside code blocks was set to `z-10`, the same z-index as the sticky page header. Because the button appeared later in the DOM, it painted on top of the header as users scrolled, causing it to bleed through the nav.

| Before | After |
|--------|--------|
|Screenshot 2026-04-23 at 05 13 25|Screenshot 2026-04-23 at 05 13 12|

---

- Reviews
- [x] automated review not possible because: visual/CSS-only fix, no logic changed
- Tests
- [x] Additional testing not necessary because: single CSS value change, verified by manual scroll test

### DIFF
--- a/.changeset/fix-copy-button-zindex.md
+++ b/.changeset/fix-copy-button-zindex.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo-docs-astro": patch
+---
+
+Fix copy code button z-index so it no longer appears above the sticky header when scrolling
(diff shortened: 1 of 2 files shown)


mattrothenberg (commented):
Could we not use `isolate` to fix this rather than muck with z-indexes?

pedromenezes1:
> Could we not use `isolate` to fix this rather than muck with z-indexes?

@mattrothenberg this made me question whether we need `z-index` here at all. turns out we don't. it works just fine without it, so i removed it entirely.

mattrothenberg (approved):

PR #460, opened 2026-04-30
fix(input-group): add !mb-0 to container className

mattrothenberg:
## Summary

We missed adding an instance of `!mb-0` to the `InputGroup` container className in `packages/kumo/src/components/input-group/input-group.tsx`. Currently only the standalone `<label>` container variant (line 296) explicitly applies `!mb-0`; the `<div>`-based container variants (used when a `label` prop is provided, in hybrid mode, and in individual mode) inherit margin from parent label styles.

This adds `!mb-0` to `containerClassName` itself so all container variants reset bottom margin consistently.

- Reviews
- [x] automated review not possible because: trivial single-line CSS class addition
- Tests
- [x] Additional testing not necessary because: visual-only change matching existing `!mb-0` usage on the standalone `<label>` container variant

### DIFF
--- a/.changeset/input-group-container-mb-0.md
+++ b/.changeset/input-group-container-mb-0.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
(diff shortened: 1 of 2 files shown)


mattrothenberg:
/bonk please review and approve this PR

PR #467, opened 2026-05-01
fix(docs): use onClick instead of onSelect on DropdownMenu.Item

mattrothenberg:
## Summary

- **Fix incorrect `onSelect` → `onClick`** on `DropdownMenu.Item` in `CloudflareLogoDemo.tsx`. `DropdownMenu.Item` does not have an `onSelect` prop — the only explicitly typed click handler from Base UI's `MenuItem` is `onClick`. `onSelect` only appeared to work because it's a native DOM event (`GlobalEventHandlers.onselect`) that React passes through to the underlying `<div>`.
- **Add `DropdownOnClickDemo`** to the canonical Dropdown demos showing the correct `onClick` pattern, so AI agents and developers have a clear reference.
- **Fix registry codegen demo lookup** to fall back to PascalCase directory name when the component export name differs from the demo file name (e.g. `DropdownMenu` vs `Dropdown`). This was silently dropping all 6 Dropdown demos from the AI registry.

## Details

The `CloudflareLogoDemo` used `onSelect` on 4 `DropdownMenu.Item` instances. Since that demo feeds into the registry codegen pipeline, the bad pattern propagated into `component-registry.md` and `component-registry.json` — the files AI agents consume for API guidance.

The registry codegen also had a name mismatch bug: demo metadata keys are derived from filenames (`DropdownDemo.tsx` → `"Dropdown"`) but the registry looks up by component export name (`"DropdownMenu"`). The fallback to `toPascalCase(config.dirName)` fixes this for all affected components.

- Reviews
  - [x] automated review not possible because: involves codegen pipeline behavior and demo correctness verification
- Tests
  - [x] Additional testing not necessary because: verified by running `codegen:registry --no-cache` and confirming DropdownMenu examples went from 0 → 6, all `onSelect` on dropdown items removed from registry output

### DIFF
--- a/.changeset/fix-dropdown-onclick.md
+++ b/.changeset/fix-dropdown-onclick.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Fix registry codegen to match demo examples when component export name differs from directory name (e.g. DropdownMenu vs dropdown). This restores missing examples for DropdownMenu and other affected components.
--- a/packages/kumo-docs-astro/src/components/demos/CloudflareLogoDemo.tsx
+++ b/packages/kumo-docs-astro/src/components/demos/CloudflareLogoDemo.tsx
@@ -82,7 +82,7 @@ export function CloudflareLogoCopyDemo() {
         <DropdownMenu.Content>
           <DropdownMenu.Item
             icon={CloudIcon}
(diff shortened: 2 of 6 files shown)


mattrothenberg:
/bonk can you approve the PR too?

PR #469, opened 2026-05-01
fix: gracefully fallback on invalid variant props instead of crashing

mattrothenberg:
## Summary

- Adds a shared `resolveVariant()` utility (`src/utils/resolve-variant.ts`) that safely resolves variant config objects with fallback to the default when an unknown key is passed, plus a dev-only `console.warn`
- Updates all 25 component variant functions to use it — previously 22 of 25 would throw `TypeError: Cannot read properties of undefined (reading 'classes')` on invalid variant values, while 3 (Badge, Code, Switch) had ad-hoc fallback logic
- Also fixes body-level lookups in Banner (`iconClasses`), Toast (`icon`), ClipboardText (`buttonSize`), Text (component body), and Table (`stickyColumnClasses`)
- Adds 34 regression tests: 6 for the utility itself, 28 exercising every variant function with invalid input

No default behavior changed — verified that the 3 previously-safe components used identical fallback keys.

- Reviews
- [x] automated review not possible because: cross-cutting change touching 25 components needs human review
- Tests
- [x] Tests included/updated

### DIFF
--- a/.changeset/safe-variant-fallback.md
+++ b/.changeset/safe-variant-fallback.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Gracefully fall back to default variant instead of crashing when an invalid variant prop is passed at runtime. Previously 22 of 25 components would throw `TypeError` on unknown variant values; all 25 now use a shared `resolveVariant()` utility that returns the default config and logs a dev warning.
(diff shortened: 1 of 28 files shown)


mattrothenberg:
/bonk review the approach and approve the PR

pedromenezes1 (approved):

PR #477, opened 2026-05-05
Fix Google Translate DOM mutation crash in Button

pedromenezes1:
Google Translate wraps text nodes in `<font>` elements, reparenting them from `<button>`. When Button toggles loading state, React's `insertBefore()` fails with `NotFoundError` because the text node is no longer a direct child.

Fix:
- Wrap `{children}` in `<span className="contents">`, it gives React a stable DOM reference while `display: contents` preserves layout
- Collapse Loader/Icon into a single ternary to eliminate an extra insertion point

---

- Reviews
- [x] automated review not possible because: bugfix validated by unit tests and manual reasoning about DOM behavior
- Tests
- [x] Tests included/updated

### DIFF
--- a/packages/kumo/src/components/button/button.tsx
+++ b/packages/kumo/src/components/button/button.tsx
@@ -136,10 +136,27 @@ export function buttonVariants({
     // Disabled state
     "disabled:cursor-not-allowed disabled:text-kumo-subtle",
     // Apply variant, size, shape styles from KUMO_BUTTON_VARIANTS
-    resolveVariant(KUMO_BUTTON_VARIANTS.variant, variant, KUMO_BUTTON_DEFAULT_VARIANTS.variant).classes,
-    resolveVariant(KUMO_BUTTON_VARIANTS.size, size, KUMO_BUTTON_DEFAULT_VARIANTS.size).classes,
-    resolveVariant(KUMO_BUTTON_VARIANTS.shape, shape, KUMO_BUTTON_DEFAULT_VARIANTS.shape).classes,
-    isCompactShape && resolveVariant(KUMO_BUTTON_VARIANTS.compactSize, size, KUMO_BUTTON_DEFAULT_VARIANTS.size).classes,
+    resolveVariant(
+      KUMO_BUTTON_VARIANTS.variant,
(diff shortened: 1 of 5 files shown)


packages/kumo/src/components/button/button.tsx
```diff
@@ -253,10 +270,12 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
...
-        {loading && <Loader size={size === "lg" ? 16 : 14} />}
-        {!loading && renderIconNode(IconComponent)}
-
-        {children}
+        {loading ? (
+          <Loader size={size === "lg" ? 16 : 14} />
+        ) : (
+          renderIconNode(IconComponent)
+        )}
+        {children != null && <span className="contents">{children}</span>}
```
mattrothenberg, line 278: Why `children != null` (instead of a shallow falsy check) and why `contents` which could expose this node to global CSS outside of our control?

invisal: I think `contents` is correct here because we want the children to be directly influenced by the button's style.

For example, the button has `flex items-center`. In this case, the children will be properly aligned. Without `<span className="contents">{children}</span>`, the children are not influenced by the button's class name.

```
<Button>
  <span>Hello</span>
  <SomeIcon />
</Button>
```

mattrothenberg: Ahhh, I didn't realize this was a Tailwind selector. Sorry! Carry on.

pedromenezes1: `!= null` catches both `null` and `undefined` (no children). A falsy check would also filter out 0, which is a valid React child. Unlikely, but `!= null` is the idiomatic React pattern

Screenshot 2026-05-05 at 12 20 05

pedromenezes1: `display: contents` basically makes the `<span>` invisible to layout. No box generated, so its children just slot into the button's flexbox like the span isn't there. The span only exists in the DOM so React has a stable node that Google Translate won't mess with

mattrothenberg (approved):

PR #480, opened 2026-05-05
fix(select): added nativeLabel option to Field

najlaskr:
## Problem
The existing `Select` component uses a `Field` wrapper to render its label text to have a consistent label/description/error layout. By default, `Field` renders a native `<label>` that's associated with the control, so hovers on the label triggers the input's hover state (i.e. in `Select`).

https://github.com/user-attachments/assets/1c9e347a-58cd-4ad3-a8dd-cbe11dde2877

## Solution
Base UI's `Field.Label` takes a `nativeLabel` prop that defaults to `true`, which renders a native `<label>` element linked to the control. A `nativeLabel` prop was added to `Field` and set `nativeLabel={false}` in `Select` that tells Base UI to render a non-`<label>` element with `aria-labelledby` to avoid hover coupling while still keeping its accessible name.

---

- Reviews
- Tests
  - [ ] Tests included/updated
  - [x] Automated tests not possible - manual testing has been completed: ran `pnpm lint` and `pnpm typecheck` and visual tests across docs

### DIFF
--- a/.changeset/select-hover-state.md
+++ b/.changeset/select-hover-state.md
@@ -1,0 +1,7 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Add a `labelNative` option to `Field` so consumers can opt out of rendering a native `<label>` wrapper when needed.
+
+Update `Select` to pass `labelNative={false}` to avoid invalid nested label semantics while preserving label rendering.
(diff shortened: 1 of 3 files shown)


najlaskr:
will create new PR with breaking API change

PR #498, opened 2026-05-07
Deprecate Link `to` prop — standardize on `href` + LinkProvider

mattrothenberg:
## Summary

- Deprecates the `to` prop on `Link` and `LinkComponentProps` in favor of `href`
- Adds a dev-mode deprecation warning when `to` is used
- Documents the correct `LinkProvider` integration pattern for framework routers
- Adds comprehensive tests covering `href`, `LinkProvider`, `render` prop, and the deprecation

## Problem

Kumo's `Link` component accepts both `href` and `to` for specifying the destination URL. This causes a real-world bug when applications configure a `LinkProvider` with a client-side router:

1. Developer writes `<Link href="https://example.com">` (the natural, platform-standard prop)
2. The `LinkProvider` wraps a router component (e.g. React Router's `<Link>`) that reads `to` for navigation
3. The router component ignores `href` — the external link silently breaks, navigating to a broken internal route instead

The workaround has been to use `<Link to="https://example.com">` instead, but **this doesn't actually work either** — React Router's `<Link to>` does not support fully qualified URLs. It resolves everything as a path within the app.

### The full matrix

**With a LinkProvider wrapping React Router:**

| URL | Prop | Result |
|---|---|---|
| `/dashboard` | `href` | Broken — RR ignores `href`, reads `to` (undefined) |
| `/dashboard` | `to` | Works — RR reads `to`, client-side nav |
| `https://example.com` | `href` | Broken — RR ignores `href` |
| `https://example.com` | `to` | **Also broken** — RR resolves it as a relative path |

No single prop solves external links without the wrapper being explicit about it.

### Why `to` exists today

`to` is a React Router-ism that was added to `LinkComponentProps` so the `DefaultLinkComponent` could accept either prop. This leaked a routing-framework concept into a presentational component's API.

### The scope of the workaround inside Kumo

Five places inside Kumo itself defensively pass `to={href}` to work around this:

| Component | Pattern |
|---|---|
| `LinkButton` | `to={typeof href === "string" ? href : undefined}` |
| `Sidebar.MenuButton` (2 sites) | `to={href}` |
| `DropdownMenu.LinkItem` | `to={href}` |
| `Breadcrumbs.Link` | `to={href}` |

### 1. Deprecate `to`, standardize on `href` (this PR)

**Link should only speak `href`** (the platform primitive). Routing integration is the application's responsibility via `LinkProvider`. This is [inversion of control](https://en.wikipedia.org/wiki/Inversion_of_control): Kumo defines the interface, the application injects the routing implementation.

This PR:
- **Deprecates `to`** with `@deprecated` JSDoc and a dev-mode `console.warn`
- **Documents `LinkProvider`** with concrete examples for React Router and Next.js
- **Adds tests** covering `href`, `LinkProvider`, `render` prop, backwards compatibility, and the deprecation warning

`to` is **not removed** — it continues to work for backwards compatibility. Consumers get a dev-mode warning guiding them to migrate.

### 2. Consumer apps build a smart `LinkProvider` wrapper (downstream follow-up)

The recommended pattern is for consumer apps to provide a `LinkProvider` component that:
- Maps `href` to the router's navigation prop for internal links
- Renders a plain `<a>` for external links
- Handles `target`, `rel`, etc. as appropriate

```tsx
const AppLink = forwardRef(({ href, to, ...rest }, ref) => {
  const destination = href ?? to;
  const isExternal = destination?.startsWith("http") &&
    new URL(destination).origin !== window.location.origin;

  if (isExternal) {
    return <a ref={ref} href={destination} {...rest} />;
  }
  return <RouterLink ref={ref} to={destination} {...rest} />;
});

<LinkProvider component={AppLink}>
  <App />
</LinkProvider>
```

Engineers then just use `href` everywhere — the wrapper handles the rest:

```tsx
<Link href="/dashboard">Dashboard</Link>
<Link href="https://example.com" target="_blank" rel="noopener noreferrer">
  Cloudflare Docs <Link.ExternalIcon />
</Link>
```

For exceptional cases where direct router control is needed, the `render` prop remains as an escape hatch:

```tsx
<Link render={<RouterLink to="/special" />}>Special case</Link>
```

## Breaking change?

**Soft deprecation only.** `to` continues to work — consumers just get a dev-mode warning guiding them to migrate. A future major version could remove `to` entirely.

- Reviews
  - [x] automated review not possible because: RFC for team discussion, requesting feedback on approach
- Tests
  - [x] Tests included/updated

### DIFF
--- a/.changeset/deprecate-link-to-prop.md
+++ b/.changeset/deprecate-link-to-prop.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Deprecate `to` prop on Link in favor of `href`. The `to` prop is a routing-framework concept that doesn't belong on a presentational component. Use `href` for all link destinations and configure a `LinkProvider` wrapper to bridge to your router. `to` continues to work but emits a dev-mode deprecation warning.
--- a/packages/kumo-docs-astro/src/pages/components/link.mdx
+++ b/packages/kumo-docs-astro/src/pages/components/link.mdx
@@ -90,11 +90,52 @@ export default function Example() {
 }
 ```
 
-### Framework Integration (render prop)
+### Framework Integration (LinkProvider)
 
 <p>
-  Use the `render` prop to compose Link styles onto framework-specific routing
-  components like React Router's Link or Next.js Link.
+  For app-wide router integration, configure a `LinkProvider` at your app root.
+  Your wrapper component receives `href` and is responsible for bridging to your
+  router's API. This lets engineers use `<Link href="...">` everywhere without
+  thinking about routing internals.
+</p>
+
+```tsx
+import { forwardRef } from "react";
+import { LinkProvider } from "@cloudflare/kumo";
+import { Link as RouterLink } from "react-router-dom";
+
+// Your app's wrapper maps href to the router's navigation prop
+// and handles external URLs with a plain <a>
+const AppLink = forwardRef(({ href, to, ...rest }, ref) => {
+  const destination = href ?? to;
+  const isExternal =
+    destination?.startsWith("http") &&
+    new URL(destination).origin !== window.location.origin;
+
+  if (isExternal) {
+    return <a ref={ref} href={destination} {...rest} />;
+  }
+  return <RouterLink ref={ref} to={destination} {...rest} />;
+});
+
+// Wrap your app once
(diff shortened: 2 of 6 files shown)


1000hz (approved):
LGTM!

PR #501, opened 2026-05-08
refactor(sensitive-input): rewrite on InputGroup + anchored toast

mattrothenberg:
## Summary

- Rewrites SensitiveInput as a thin wrapper around InputGroup, replacing 500+ lines of hand-rolled layout (manual mask overlay, absolute positioning, per-size padding duplication, `role="button"` container, `execCommand` fallback) with composable primitives
- Copy confirmation now uses an anchored Base UI Toast positioned above the copy button instead of icon-swapping with a timeout
- Action button tooltips use a scoped inverted dark style (`bg-kumo-contrast`/`text-kumo-base`) that adapts to light/dark mode
- Toast dismisses immediately on blur (e.g. tabbing to reveal button) for snappy feedback
- Consumer `onBlur` is properly forwarded before masking logic (verified against stratus Formik usage)
- No public API changes — all existing props (`onChange`, `onValueChange`, `onBlur`, `error`, `onCopy`, etc.) continue to work

Audited against real-world usage in `cloudflare/stratus` (CustomHeaders.tsx, ImportVariablesSecretsDialog.tsx) — no breaking changes.

- Reviews
- [x] automated review not possible because: UX refactor needs visual review
- Tests
- [x] Additional testing not necessary because: existing unit tests pass, no API changes, verified against stratus consumers

### DIFF
--- a/packages/kumo/src/components/sensitive-input/sensitive-input.tsx
+++ b/packages/kumo/src/components/sensitive-input/sensitive-input.tsx
@@ -86,6 +86,7 @@ export const SensitiveInput = forwardRef<HTMLInputElement, SensitiveInputProps>(
       value: controlledValue,
       defaultValue = "",
       onChange,
+      onBlur,
       onValueChange,
       onCopy,
       size = KUMO_SENSITIVE_INPUT_DEFAULT_VARIANTS.size,
@@ -126,6 +127,7 @@ export const SensitiveInput = forwardRef<HTMLInputElement, SensitiveInputProps>(
     // Mask on blur (unless focus moves to a sibling button inside the group)
     const handleBlur = useCallback(
       (e: React.FocusEvent<HTMLInputElement>) => {
+        onBlur?.(e);
         if (
           containerRef.current &&
           e.relatedTarget instanceof Node &&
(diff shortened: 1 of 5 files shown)


mattrothenberg:
/bonk what do you think about these changes

packages/kumo/src/components/sensitive-input/sensitive-input.tsx
```diff
@@ -1,23 +1,22 @@
-import { Eye, EyeSlash } from "@phosphor-icons/react";
+import { Copy, Eye, EyeSlash } from "@phosphor-icons/react";
```
pedromenezes1: afaik phosphor throws a warning when we import them like this `Copy` and not `CopyIcon`

mattrothenberg: okie doke, will fix

pedromenezes1:
good addition with the `labels` prop

pedromenezes1:
Screenshot 2026-05-08 at 17 26 31

I don't think I planned how it would look with two buttons in a single `InputGroup.Addon`. it might be a good opportunity to tweak it if you think `gap-1.5` leaves them too close together.

mattrothenberg:
@pedromenezes1 Yep, I think the double button will be controversial, so I wanted to put this up and get feedback :)

PR #520, opened 2026-05-18
fix(styles): show pointer cursor on clickable elements

awhobbs87:
## Summary
- Add a low-specificity `:where()` rule so clickable native and ARIA elements use the pointer cursor by default.
- Exclude disabled, `aria-disabled`, and `data-disabled` elements.
- Add a patch changeset for `@cloudflare/kumo`.

## Testing
- `pnpm exec prettier --check packages/kumo/src/styles/kumo.css .changeset/clickable-cursor.md`
- `pnpm --filter @cloudflare/kumo lint` (3 existing warnings, no errors)

- Reviews
- [x] automated review not possible because: small CSS-only affordance update
- Tests
- [ ] Tests included/updated
- [x] Additional testing not necessary because: covered by formatting, lint, and package build


packages/kumo/src/styles/kumo.css
```diff
@@ -20,6 +20,32 @@
   }
 }
 
+/* Base clickable affordances */
+:where(
+    a[href],
+    button,
+    input[type="button"],
+    input[type="submit"],
+    input[type="reset"],
```
eduardo-vargas, line 38: minor: I would include these two here as well

```css
input[type="checkbox"],
input[type="radio"],
```

geoquant (dismissed):
Thanks for putting this together — I’m supportive of the pointer affordance itself, but I’d like us to avoid scoping this via `[class*="kumo-"]` before merging.

That selector couples global behavior to our Tailwind/token class names, is broader than we need, and feels like a pattern we probably don’t want to bless for future global component styles.

Could we instead add an explicit Kumo-owned data attribute to rendered component roots/parts and scope this rule to that? For example:

```tsx
<button data-kumo-component="Button" />

<Select.Trigger
  data-kumo-component="Select"
  data-kumo-part="trigger"
/>
```

Then the CSS can target Kumo-rendered interactive elements without a star selector or class-name probing:

```css
:where(
    [data-kumo-component],
    [data-kumo-part]
  ):where(
    a[href],
    button,
    input[type="button"],
    input[type="checkbox"],
    input[type="radio"],
    input[type="submit"],
    input[type="reset"],
    label[for],
    select,
    summary,
    [role="button"],
    [role="link"],
    [role="menuitem"],
    [role="menuitemcheckbox"],
    [role="menuitemradio"],
    [role="option"],
    [role="tab"],
    [role="checkbox"],
    [role="radio"],
    [role="switch"]
  ):where(:not(:disabled)):where(:not([disabled])):where(
    :not([aria-disabled="true"])
  ):where(:not([data-disabled])) {
  cursor: pointer;
}
```

This gives us a more sustainable scoping primitive for Kumo-owned global styles beyond just this cursor rule. I’d start by adding the data attrs to the interactive components/parts this PR wants to affect, e.g. Button, Link, Checkbox, Radio, Switch, Select trigger/options, DropdownMenu items/triggers, Popover/Tooltip/Dialog triggers, Tabs, Pagination, Breadcrumb links, Collapsible trigger, etc.

awhobbs87:
Thanks @geoquant — adopted in dda624119. Swapped the `[class*="kumo-"]` scoping for `[data-kumo-component]` / `[data-kumo-part]` attributes, and threaded the attributes through all the interactive component roots/parts you called out (Button, Link, Checkbox, Radio, Switch, Select trigger/option, DropdownMenu items/triggers, Dialog/Popover triggers, Tabs, Collapsible, Breadcrumbs, TableOfContents, Sidebar, MenuBar, Combobox, Autocomplete, Toast close, SensitiveInput).

Skipped a few intentionally:
- Tooltip trigger (uses `cursor-default` intentionally; a tooltip trigger is a disclosure, not an action)
- Table column resize handle (`cursor-col-resize`)
- DatePicker day buttons (already covered by `.rdp-day_button { cursor: pointer }` in the same stylesheet)
- date-range-picker (deprecated, already sets `cursor-pointer` inline)

All 1004 unit tests pass; typecheck, lint, and build all clean.

PR #530, opened 2026-05-21
fix: constrain Tooltip popup width to available viewport space

mattrothenberg:
## Problem

Kumo's `Tooltip` popup has no `max-width` constraint. When tooltip content is long (e.g. multi-sentence descriptions), the popup grows unbounded and overflows past the viewport edge.

## Fix

Adds `max-w-[var(--available-width)]` to the Tooltip popup. `--available-width` is a CSS variable that Base UI's `Positioner` already sets on the positioner element — it represents the available space between the trigger and the viewport edge. Since CSS custom properties inherit, the Popup (a child of the Positioner) can use it directly.

This means:
- Tooltips near the center of the page remain unconstrained (available width is large)
- Tooltips near viewport edges automatically wrap to fit
- No hardcoded magic number — the constraint is dynamic and responsive

Applied to both the `Tooltip` component and the `tooltipVariants()` utility function.

## Demo

Added a "Long Content / Overflow" example to the Astro docs (`/components/tooltip`) with three triggers spread across the full width, each with long lorem ipsum content. Hovering the edge triggers shows the tooltip wrapping instead of overflowing.

## Checklist

- [x] automated review not possible because: single CSS class addition, no logic changes
- [x] Additional testing not necessary because: visual behavior only — added a demo to the Astro docs for manual verification

### DIFF
--- a/.changeset/tooltip-overflow-constraint.md
+++ b/.changeset/tooltip-overflow-constraint.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Fix Tooltip popup overflowing viewport when content is wider than available space. The popup now constrains its width to `var(--available-width)`, a CSS variable provided by Base UI's Positioner that reflects the space between the trigger and the viewport edge.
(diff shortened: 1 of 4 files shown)


mattrothenberg:
/bonk please review and approve these changes

PR #535, opened 2026-05-22
feat(kumo): add disabled prop to LinkButton

andystalick:
## Summary

Adds a `disabled?: boolean` prop to `LinkButton`. When disabled, `LinkButton` renders a real, natively-disabled `<button>` by delegating to the `Button` component — rather than keeping an `<a>` and marking it `aria-disabled`.

> This replaces the earlier `aria-disabled` approach from this PR's history with a simpler, more robust implementation.

## Behavior when `disabled` is `true`

- Renders `<button disabled type="button" data-kumo-component="LinkButton">` (via the `Button` component)
- Anchor-only attributes (`href`, `target`, `rel`, `download`, `hrefLang`, `media`, `ping`, `referrerPolicy`, `linksExternal`) are stripped — they're invalid on a `<button>`
- All `on*` event handlers are dropped — they're inert on a disabled control
- `title` is forwarded to `Button`, which renders it as a Kumo `Tooltip` (useful for explaining *why* the action is unavailable)
- Same disabled styling as `Button`, since it *is* a `Button`

## Why render as `<button>` instead of `aria-disabled` on `<a>`

A disabled link is not a link — there is no valid `href` to navigate to. Native `<button disabled>` is inert for free: the browser blocks clicks, keyboard activation, and focus without any custom `preventDefault`/`stopPropagation` handlers, and assistive tech announces it correctly. Delegating to `Button` also means we inherit its existing disabled styling and title→tooltip behavior with almost no new code, instead of maintaining a parallel `aria-disabled:` styling matrix.

Trade-off: `ref` (typed `HTMLAnchorElement`) is intentionally **not** forwarded in the disabled branch, since the rendered element is a `<button>`. This is documented with an inline comment.

## Implementation notes

- A small `toDisabledButtonProps()` helper strips anchor-only attributes and event handlers, returning `React.ButtonHTMLAttributes<HTMLButtonElement>` behind a single contained cast.
- `shape` is passed to `Button` as `shape as "base"` because `Button`'s discriminated union (`ButtonWithTextProps | IconOnlyButtonProps`) rejects a dynamically-typed shape. Safe at runtime (the value only feeds `buttonVariants`), and no a11y regression since `LinkButton` never enforced icon-only `aria-label`.

## Tests

4 new tests in `button.test.tsx` under a `LinkButton > disabled` block:
- Renders a disabled `<button>` (not an `<a>`), with `type="button"`
- Strips anchor-only attributes (`href`, `target`, `rel`, `download`) from the rendered button
- Preserves `data-kumo-component="LinkButton"`
- Wraps `title` in an enabled tooltip trigger around the disabled button

All 23 `button.test.tsx` tests pass.

## Docs

- New `ButtonDisabledLinkDemo` in `ButtonDemo.tsx` (disabled link + disabled-with-tooltip)
- New "Disabled Link" subsection on `/components/button`
- JSDoc on `LinkButtonProps.disabled` documents the behavior for IDE tooltips

## Repo cleanup (bundled in)

- Removed stale Storybook references from `packages/kumo/README.md`
- Dropped `.stories.tsx` generation from the component scaffolder (`plopfile.js` + deleted the story template), since Storybook is no longer set up in this repo

## Verification

- `pnpm --filter @cloudflare/kumo typecheck` → 0 errors
- `pnpm lint` → 0 errors (7 pre-existing warnings in chart code, unrelated)
- `pnpm --filter @cloudflare/kumo test:run` → 1202/1204 pass; the 2 failures are pre-existing `./components/toolbar` export-path checks against a stale `dist/` and are unrelated to this change
- Manual browser smoke test on `/components/button`: enabled state renders `<a href>`; disabled state renders `<button disabled type="button">` with `href` dropped and `data-kumo-component="LinkButton"` preserved; disabled styling correct; `title` tooltip fires on hover

## Changeset

`minor` bump for `@cloudflare/kumo` (additive prop, no breaking change).

---

- Reviews
- [x] automated review not possible because: feature work on a personal fork — bonk reviews can run after merge to main
- Tests
- [x] Tests included/updated


packages/kumo/src/components/button/button.tsx
```diff
@@ -133,8 +133,9 @@ export function buttonVariants({
     "border-0 shadow-xs",
     "focus:outline-none focus:ring-kumo-focus/50 focus-visible:ring-2 focus-visible:ring-kumo-brand",
     "cursor-pointer",
-    // Disabled state
+    // Disabled state — applies to native :disabled (Button) and aria-disabled (LinkButton)
     "disabled:cursor-not-allowed disabled:text-kumo-subtle",
+    "aria-disabled:cursor-not-allowed aria-disabled:text-kumo-subtle aria-disabled:pointer-events-none",
```
andystalick: We may roll back pointer-events-none and allow the JS to handle things.

packages/kumo/src/components/button/button.tsx
```diff
@@ -327,12 +338,15 @@ export const LinkButton = React.forwardRef<HTMLAnchorElement, LinkButtonProps>(
     {
       children,
       className,
+      disabled = false,
       external,
       href,
       shape = "base",
       size = "base",
       variant = "ghost",
```
andystalick, line 427: This would be a breaking change, but why is the default `ghost` here when `Button`'s default is `secondary`?

stritt:
thanks for this! the disabled linkbutton support still looks useful. this mainly needs a rebase, the open review threads around pointer events and the default variant resolved and a fresh ci run, then we can get it reviewed again.

andystalick:
Alright, I went ahead and re-implemented this based on what I've learned.

packages/kumo/src/components/button/button.tsx
andystalick: Leaving this for another day

packages/kumo/src/components/button/button.tsx
andystalick: Not applicable with the new approach

mattrothenberg (approved):

PR #537, opened 2026-05-26
feat(chart): React-native tooltip for TimeseriesChart

Ankcorn:
## Why

The existing tooltip was built on ECharts' HTML-string formatter — inline styles, `any` types everywhere, `getComputedStyle` called on every mouse move to read CSS variables, and a dead `dimColor` variable that was computed but never used. It also couldn't escape its containing element, so tooltips were clipped when the chart was inside a constrained layout.

## What changed

image
image

**Tooltip is now a proper React component** rendered via `createPortal` into `document.body`, wrapped in a `data-mode` div so Kumo CSS variables resolve correctly despite being outside the component tree.

**Positioning is decoupled from content** (same architecture as ECharts' own implementation):
- *Content* flows through React state via the `updateAxisPointer` event — one render cycle behind, imperceptible for text
- *Position* is driven by direct DOM mutation in a native `mousemove` listener using `transform: translate3d` (GPU-composited, no layout cost). Quadrant-based anchoring flips the tooltip side when the cursor crosses the chart midpoint, with a `cubic-bezier(0.23,1,0.32,1)` transition on switches.

**New props:**
- `tooltipMode`: `"all"` (default) | `"single"` — single mode shows only the series nearest the cursor, useful for dense charts
- `tooltipMaxItems`: caps rows in `"all"` mode with a `+N more` footer (default `10`)

**Other improvements:**
- Values sorted descending so the highest series appears first
- Date formatted as `May 26, 13:04:21` instead of ISO string
- Fallback value formatter uses `toLocaleString` — no scientific notation
- `findNearest` binary search replaces ECharts series index lookup

## Alternatives considered

Keeping ECharts' formatter but injecting computed CSS variable values — simpler but still `getComputedStyle` on every hover, still HTML strings, still no real types.

---
- Reviews
- [x] automated review not possible because: tooltip behaviour is visual/interactive, not unit-testable without a browser
- Tests
- [x] Additional testing not necessary because: covered by the new ManySeriesChartDemo in the docs site which exercises truncation, formatting, and positioning


mattrothenberg:
/bonk what do you think of these changes?

mattrothenberg:
/bonk please thoroughly review our changes here, looking for subtle API breakages, performance issues, etc

mattrothenberg (approved):

PR #543, opened 2026-05-26
fix: mark echarts peerDependency as optional so it isn't installed everytime

vaibhavshn:
Marks echarts as an optional peer dependency so it isn't installed everytime.

I saw that echarts was being installed that took about 66MB (61MB echarts + 5MB zrender - a dependency of echarts) on my disk when i installed it in a project. Since our docs say that we need to explicitly install it, wouldn't it be better to mark it as optional as well?

Let me know what you think about this, maybe not making it optional breaks kumo somewhere?

Doc: https://kumo-ui.com/charts/

---

- Reviews
  - [x] automated review not possible because: cannot run bonk myself 
- Tests
  - [ ] Tests included/updated
  - [x] Additional testing not necessary because: just marks dependency as optional for users

### DIFF
--- a/.changeset/vast-feet-allow.md
+++ b/.changeset/vast-feet-allow.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Mark echarts peer dependency as optional so it isn't installed everytime
--- a/.changeset/vast-feet-allow.md
+++ b/.changeset/vast-feet-allow.md
@@ -2,4 +2,4 @@
 "@cloudflare/kumo": patch
 ---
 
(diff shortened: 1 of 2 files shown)


mattrothenberg (changes requested):
Small changeset update please:
```md
echarts is now an optional peer dependency. If you don't use Chart components (Chart, TimeseriesChart, SankeyChart), it will no longer be installed automatically. If you do use them, run npm install echarts (as documented).
```

geoquant (approved):

mattrothenberg (approved):

PR #545, opened 2026-05-27
fix(sidebar): Last minute tweaks and nits

hkfoster:
## Summary

Fixes several visual and interaction issues in the collapsed sidebar:

- **Tooltip stability**: Always wrap `MenuButton` in Tooltip when `tooltip` prop is provided, keeping DOM structure stable across expand/collapse to preserve CSS transitions. Pass `content={null}` when tooltip should be suppressed (expanded state or peekable enabled).
- **Peekable tooltip suppression**: Added `peekable` to sidebar context so `MenuButton` can disable tooltips when peeking is enabled (tooltips are redundant when hovering reveals the full sidebar).
- **Explicit transition properties**: Replaced `transition-all` with `transition-[color,background-color,box-shadow,outline]` on `MenuButton` per Web Interface Guidelines.
- **Smooth ring transition**: Changed `ring-0` to `ring-transparent` on search button demo so the ring animates out instead of snapping.
- **Overflow-based text clipping**: Removed `hidden` classes from text spans in demos, relying on `overflow-hidden` + `min-w-0` for smooth width-driven clipping during collapse.
- **Caret icon transition**: Wrapped `CaretUpDownIcon` in a span with `transition-[width]` for smooth hide/show in AccountSwitcher.
- **Sliding views fix**: Removed `invisible` from inactive `SlidingView` panels — it was hiding the outgoing view instantly mid-slide. Parent `overflow-hidden` clips off-screen views; `aria-hidden` + `inert` handle accessibility.
- **Resize handle styling**: Updated to `after:w-0.5` with `bg-kumo-hairline` tokens.
- **Stale comment cleanup**: Removed outdated comment about avoiding transitions.

---

- Reviews
- [x] automated review not possible because: visual polish fixes requiring manual verification
- Tests
- [ ] Tests included/updated
- [x] Additional testing not necessary because: changes are CSS/styling only with no behavioral changes; existing tests pass (2 pre-existing inert failures unrelated)

### DIFF
--- a/packages/kumo/.oxlintrc.json
+++ b/packages/kumo/.oxlintrc.json
@@ -34,7 +34,7 @@
     // https://github.com/oxc-project/oxc/issues/1141
     "jsx-a11y-js/interactive-supports-focus": "error",
     "jsx-a11y-js/no-interactive-element-to-noninteractive-role": "error",
-    "jsx-a11y-js/no-noninteractive-element-interactions": "error",
+    "jsx-a11y-js/no-noninteractive-element-interactions": "warn",
     "jsx-a11y-js/no-noninteractive-element-to-interactive-role": "error",
     "jsx-a11y-js/control-has-associated-label": "warn"
   }
--- a/packages/kumo/.oxlintrc.json
+++ b/packages/kumo/.oxlintrc.json
@@ -34,7 +34,7 @@
     // https://github.com/oxc-project/oxc/issues/1141
     "jsx-a11y-js/interactive-supports-focus": "error",
     "jsx-a11y-js/no-interactive-element-to-noninteractive-role": "error",
-    "jsx-a11y-js/no-noninteractive-element-interactions": "warn",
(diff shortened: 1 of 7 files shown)


packages/kumo/.oxlintrc.json
```diff
@@ -34,7 +34,7 @@
     // https://github.com/oxc-project/oxc/issues/1141
     "jsx-a11y-js/interactive-supports-focus": "error",
     "jsx-a11y-js/no-interactive-element-to-noninteractive-role": "error",
-    "jsx-a11y-js/no-noninteractive-element-interactions": "error",
+    "jsx-a11y-js/no-noninteractive-element-interactions": "warn",
```
Brayden: Is this change intentional?

hkfoster: It was, but after seeing how shadcn handles their sidebar resizing button, I reverted it. Should be good to go.

PR #549, opened 2026-05-28
fix(docs): fix timeseries page TOC and loading state hydration

mattrothenberg:
Fixes two issues on the timeseries chart docs page:

**Empty Table of Contents**
All section headings were `###` (h3) with no `##` (h2) parents. The `TableOfContents` component groups h3s under h2s and silently drops orphaned h3s, so the TOC rendered empty. Promoted all headings to h2.

**Broken loading state demo**
The `LoadingChartDemo` uses `useIsDarkMode()` which reads from the DOM. During SSR, it defaults to light mode and renders the wave SVG with a dark stroke color. On client hydration in dark mode, React detects the attribute mismatch and refuses to patch it — resulting in an invisible wave (black stroke on dark background). Switched to `client:only="react"` to skip SSR entirely.

---

- Reviews
- [x] automated review not possible because: docs-only change with no component logic
- Tests
- [x] Additional testing not necessary because: docs-only MDX heading level and hydration directive change, verified locally


packages/kumo-docs-astro/src/pages/charts/timeseries.mdx
```diff
@@ -96,40 +94,7 @@ import {
...
-</ComponentSection>
-
-<ComponentSection>
-
-### Tooltip Boundary
-
-<p>
-  Use the <code>tooltipBoundary</code> prop to constrain the tooltip to a
-  specific container element. By default the tooltip avoids overflowing any
-  clipping ancestor (scroll containers, viewports). Pass a DOM element to
```
mattrothenberg, line 120: This example wasn't rendering at all

PR #551, opened 2026-05-29
Expose useFilter from Combobox and Autocomplete

pedromenezes1:
Re-export Base UI's `useFilter` from Kumo's `Combobox` and `Autocomplete` namespaces.

Base UI defaults to case- and accent-insensitive filtering via `Intl.Collator`, but consumers writing a custom `filter` often silently override it with naive `.toLowerCase().includes()`. Typing "portugues" then fails to find "Português", and "cote" fails to find "Côte d'Ivoire". Two of our own demos shipped this anti-pattern.

- Re-export `useFilter` and `Filter` types from `Combobox` and `Autocomplete`
- Fix the two broken Autocomplete demos to use `useFilter().contains`
- Switch combobox demos to a shared 12-item native-name `languages` list so accent-insensitivity is demonstrable
- Document the default + override pattern in `combobox.mdx` and `autocomplete.mdx`

After the change, the recommended pattern is:
```tsx
function LanguagePicker() {
  const { contains } = Autocomplete.useFilter();
  return (
    <Autocomplete
      items={languages}
      filter={(item, query) => contains(item.label, query)}
    >
      {/* ... */}
    </Autocomplete>
  );
}
```

---

- Reviews
- [x] automated review not possible because: mostly demo data, docs, and a single re-export
- Tests
- [ ] Tests included/updated
- [ ] Automated tests not possible - manual testing has been completed as follows: <description>
- [x] Additional testing not necessary because: re-exports an already-tested Base UI hook

### DIFF
--- a/packages/kumo-docs-astro/src/components/demos/AutocompleteDemo.tsx
+++ b/packages/kumo-docs-astro/src/components/demos/AutocompleteDemo.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useCallback, useState } from "react";
 import { Autocomplete } from "@cloudflare/kumo";
 import { languages, type Language } from "./data/languages";
 
@@ -125,15 +125,18 @@ export function AutocompleteDemo() {
 export function AutocompleteWithFieldDemo() {
   const { contains } = Autocomplete.useFilter();
 
+  const filter = useCallback(
+    (item: Language, query: string) => contains(item.label, query),
+    [contains],
+  );
+
   return (
     <div className="w-80">
       <Autocomplete
         items={languages}
         label="Language"
         description="Start typing to filter languages"
-        filter={(item: Language, query: string) =>
-          contains(item.label, query)
-        }
+        filter={filter}
       >
         <Autocomplete.InputGroup placeholder="Search a language…" />
         <Autocomplete.Content>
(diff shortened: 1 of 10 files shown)


packages/kumo-docs-astro/src/components/demos/AutocompleteDemo.tsx
```diff
@@ -122,22 +123,24 @@ export function AutocompleteDemo() {
...
-        items={countries}
-        label="Country"
-        description="Start typing to filter countries"
-        filter={(item: Country, query: string) =>
-          item.label.toLowerCase().includes(query.toLowerCase())
+        items={languages}
+        label="Language"
+        description="Start typing to filter languages"
+        filter={(item: Language, query: string) =>
+          contains(item.label, query)
```
mattrothenberg: Why are these type annotations necessary? Is this not inferred from the `items` prop?

packages/kumo-docs-astro/src/components/demos/AutocompleteDemo.tsx
```diff
@@ -149,14 +152,16 @@ export function AutocompleteWithFieldDemo() {
...
+
   return (
     <div className="w-80">
       <Autocomplete
         items={countries}
         label="Country"
         error={{ message: "Please enter a valid country", match: true }}
         filter={(item: Country, query: string) =>
-          item.label.toLowerCase().includes(query.toLowerCase())
+          contains(item.label, query)
```
mattrothenberg: If we're aiming for correctness in our demos/examples, this should probably go up a layer and be wrapped in `useCallback` so it's not re-created on every render.

mattrothenberg (approved):

packages/kumo-docs-astro/src/components/demos/AutocompleteDemo.tsx
pedromenezes1: good catch. pushed an update with both fixes.

packages/kumo-docs-astro/src/components/demos/AutocompleteDemo.tsx
pedromenezes1: agreed. pushed.

PR #560, opened 2026-05-30
feat(dialog): add `busy` prop, `size="2xl"`, `verticalAlign` + `topOffset`

awhobbs87:
## Summary

Three additive, non-breaking improvements to `Dialog`:

- **`<Dialog.Root busy>`** - dynamic flag that blocks user-initiated dismissal while an async action is in flight.
- **`<Dialog size="2xl">`** - new size variant (min 1024px) for data-dense layouts.
- **`<Dialog verticalAlign="top" topOffset={N}>`** - anchor the dialog to the top of the viewport with a configurable pixel offset, instead of vertical centring.

Each of these was a paper cut hit repeatedly by a downstream Cloudflare product that had to fork the dialog primitive to work around the gaps. This PR upstreams the missing props so the fork can go away.

### `busy`

Today `Dialog.Root` exposes `disablePointerDismissal` for blocking outside clicks, but it is a static prop and only covers one of the four dismissal paths. Real-world Save dialogs need to block all four (Escape, outside click, focus-out, and `<Dialog.Close>` button presses) only while the async Save handler is in flight, and only for the duration of that handler.

The new `busy` prop wraps base-ui's `onOpenChange(open, details)` and swallows the call when `!open && details.reason` is one of `outside-press`, `escape-key`, `close-press`, or `focus-out`. Programmatic close (controlled `open={false}` from the parent, or `actionsRef.close()`) still works, so the dialog can be dismissed once the async work resolves. This complements `disablePointerDismissal`, it does not replace it.

### `size="2xl"`

Existing sizes top out at `xl` = 768px, which is too narrow for dialogs containing two columns of form fields, a summary table, or a side-by-side preview. `2xl` adds a 1024px minimum width using the same `min-w-[Nrem]` Tailwind pattern as the other sizes, and is wired into both `KUMO_DIALOG_VARIANTS.size` and `KUMO_DIALOG_STYLING.dimensions` so the Figma plugin metadata stays consistent.

### `verticalAlign` + `topOffset`

Default vertical centring competes visually with persistent app navbars. Top-anchored dialogs with a configurable navbar-clearing offset are a common requirement. Default behaviour (`verticalAlign="center"`) is unchanged; opt-in `verticalAlign="top" topOffset={N}` flips the panel to inline `top: Npx` and drops the `-translate-y-1/2` centring class.

## API

```tsx
// busy: pure-additive on Dialog.Root, works for both role="dialog" and role="alertdialog"
<Dialog.Root open={open} onOpenChange={setOpen} busy={saving}>
  <Dialog>
    <Dialog.Title>Save changes</Dialog.Title>
    <Button onClick={async () => { setSaving(true); await save(); setSaving(false); setOpen(false); }}>
      Save
    </Button>
  </Dialog>
</Dialog.Root>

// size="2xl": new variant slot
<Dialog size="2xl">{/* min-w 1024px */}</Dialog>

// verticalAlign + topOffset: new content-panel props
<Dialog verticalAlign="top" topOffset={64}>
  {/* anchored 64px from the top of the viewport */}
</Dialog>
```

## Tests

Adds `packages/kumo/src/components/dialog/dialog.test.tsx` with 10 tests (the file did not exist before, so this is also the first test file for the Dialog component):

- baseline render (title, description, close button visible)
- `busy` swallows close-button presses
- `busy` swallows Escape
- `busy=false` allows close-button presses (control)
- `busy` swallows Escape on `role="alertdialog"`
- programmatic close still works while `busy=true` (regression guard against making the API a one-way street)
- `size="2xl"` applies the `min-w-[64rem]` class
- default `verticalAlign="center"` applies `top-1/2` + `-translate-y-1/2`
- `verticalAlign="top" topOffset={49}` applies inline `top: 49px` and drops the centring classes
- `verticalAlign="top"` without `topOffset` defaults to `top: 0px`

All 10 pass. Full kumo suite: 1,120/1,120 pass. Typecheck clean. Lint clean (0 errors). Prettier clean.

## Docs

- Three new demos in `kumo-docs-astro/src/components/demos/DialogDemo.tsx`: `DialogBusyDemo`, `DialogSize2xlDemo`, `DialogVerticalAlignTopDemo`.
- Three new Examples sections in `kumo-docs-astro/src/pages/components/dialog.mdx`: "Busy state", "Wide layout", "Top-anchored".
- `busy` row added to the manual Dialog.Root props table.
- `size="2xl"`, `verticalAlign`, and `topOffset` flow through the auto-generated `PropsTable` via JSDoc on the prop types.

## Screenshots

**`busy=true` mid-Save** - Escape, outside clicks, and close button are all temporarily disabled:

_(drag busy-saving.png here)_

**`size="2xl"`** - 1024px min-width gives room for two-column layouts:

_(drag size-2xl.png here)_

**`verticalAlign="top" topOffset={64}`** - anchored to top of viewport, navbar-clearing gap visible:

_(drag vertical-align-top.png here)_

## Compatibility

- 100% backwards compatible. All new props are optional with defaults that preserve current behaviour:
  - `size` default unchanged (`"base"`).
  - `verticalAlign` defaults to `"center"`, which expands to the same `top-1/2 left-1/2 -translate-x-1/2 -translate-y-1/2` style stack as before.
  - `busy` defaults to `false`, in which case `onOpenChange` is passed through to base-ui untouched.
- The `dialogVariants()` cn string was split: the centring classes moved out of the always-applied base into the `verticalAlign.center` variant slot, but the rendered class set for an unparameterised `<Dialog>` is byte-identical to before (verified by the baseline render test).

## Changeset

`minor`, matching the precedent set by `feat(chart): React-native tooltip for TimeseriesChart` (#537) where net-new public surface used `minor` rather than `patch`.

## Notes for reviewers

- The `DISMISSAL_REASONS` set uses hardcoded string literals rather than importing from `@base-ui/react/internals/reasons` because the `internals` path is marked private by base-ui. The values are part of the stable public surface via the `DialogRootChangeEventReason` type union, but the constants live in an internal module. Tests cover all four reasons (well, two of the four directly, plus the alertdialog variant) to catch drift if base-ui ever renames one.
- The `verticalAlign="top"` `topOffset` is implemented via inline `style.top` rather than a Tailwind utility because the offset value is arbitrary (could be `64`, `49`, `80`, etc. per app's navbar). This matches the existing inline-style approach for the dialog's `transition-property` and `--tw-shadow`.
- Did not add `outside-press` or `focus-out` test coverage because jsdom makes both awkward to trigger deterministically. The set-based reason check is symmetric across all four values, so escape/close-press passing is good evidence that the other two also work.

### DIFF
--- a/.changeset/dialog-busy-size-position.md
+++ b/.changeset/dialog-busy-size-position.md
@@ -1,0 +1,11 @@
+---
+"@cloudflare/kumo": minor
+---
+
+**Dialog: `busy` prop, `size="2xl"`, and `verticalAlign="top"` with `topOffset`**
+
+Three additive Dialog improvements, all non-breaking:
+
+- **`<Dialog.Root busy>`**: when true, blocks user-initiated dismissal (Escape key, outside click, focus-out, and `<Dialog.Close>` button presses) so an in-flight async action can't be abandoned by a stray keystroke or backdrop click. Programmatic close (controlled `open={false}` or `actionsRef.close()`) still works, so the dialog can be dismissed once the async work resolves. Designed to flip dynamically while a Save handler is in flight, complementing the existing static `disablePointerDismissal`.
+- **`<Dialog size="2xl">`**: new size variant (min 1024px) for data-dense layouts that need more horizontal real estate than `xl` (768px) provides — side-by-side forms, summary tables, multi-pane previews.
+- **`<Dialog verticalAlign="top" topOffset={N}>`**: anchor the dialog to the top of the viewport with a configurable pixel offset, instead of the default vertical centring. Useful for apps with a persistent navbar where a centered dialog would visually compete with the chrome.
--- a/packages/kumo-docs-astro/src/components/demos/DialogDemo.tsx
+++ b/packages/kumo-docs-astro/src/components/demos/DialogDemo.tsx
@@ -297,6 +297,177 @@ export function DialogWithComboboxDemo() {
   );
 }
 
+/**
+ * Demonstrates the `busy` prop. When `busy={true}`, Escape, outside clicks,
+ * and Dialog.Close buttons are all swallowed, preventing the user from
+ * abandoning an in-flight Save. Programmatic close still works (the demo
+ * resolves and closes after 2s).
+ */
+export function DialogBusyDemo() {
+  const [open, setOpen] = useState(false);
+  const [saving, setSaving] = useState(false);
+
+  return (
+    <Dialog.Root open={open} onOpenChange={setOpen} busy={saving}>
+      <Dialog.Trigger render={(p) => <Button {...p}>Open</Button>} />
+      <Dialog className="p-8">
+        <div className="mb-4 flex items-start justify-between gap-4">
+          <Dialog.Title className="text-2xl font-semibold">
+            {saving ? "Saving…" : "Save changes"}
+          </Dialog.Title>
+          <Dialog.Close
+            aria-label="Close"
+            render={(props) => (
+              <Button
+                {...props}
+                variant="secondary"
+                shape="square"
+                icon={<X />}
+                aria-label="Close"
+              />
+            )}
+          />
+        </div>
+        <Dialog.Description className="text-kumo-subtle">
+          {saving
+            ? "Don't go anywhere — Escape, outside clicks, and the close button are temporarily disabled while we save your changes."
+            : "Click Save to start a fake 2-second save. While it runs, Escape, outside clicks, and Close are blocked."}
+        </Dialog.Description>
+        <div className="mt-8 flex justify-end gap-2">
+          <Dialog.Close
+            render={(props) => (
+              <Button variant="secondary" {...props} disabled={saving}>
+                Cancel
+              </Button>
+            )}
+          />
+          <Button
+            variant="primary"
+            disabled={saving}
+            onClick={() => {
+              setSaving(true);
+              setTimeout(() => {
+                setSaving(false);
+                setOpen(false);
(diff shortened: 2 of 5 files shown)


mattrothenberg:
Thanks for putting this together. I agree these are real downstream paper cuts, but I don’t think we should solve them by adding this much new Dialog API surface.

The `busy` prop is the clearest non-starter for me. Blocking dismissal while async work is in flight is application policy, not Dialog primitive behavior. It also creates a second dismissal-control abstraction alongside Base UI’s existing `onOpenChange`/controlled state model and our `disablePointerDismissal`. I’d prefer downstream apps handle that in their own controlled wrapper.

For the width and positioning use cases, I think the existing `className` and `style` escape hatches already solve this without new public props:

```tsx
// Wide dialog
<Dialog className="sm:min-w-[64rem]">
  ...
</Dialog>

// Top-anchored with static offset
<Dialog className="top-[64px] translate-y-0">
  ...
</Dialog>

// Top-anchored with dynamic offset
<Dialog className="translate-y-0" style={{ top: headerHeight }}>
  ...
</Dialog>
```

I tested these locally in the docs and they work as expected because `Dialog` merges caller `className` after `dialogVariants(...)` through `cn()`, so `tailwind-merge` lets caller classes override `top-1/2` and `-translate-y-1/2`.

My recommendation would be to remove `busy`, `size="2xl"`, `verticalAlign`, and `topOffset` from this PR. If we want to improve discoverability, let’s add docs examples showing these `className`/`style` patterns instead. If `2xl` later becomes a canonical design-system size, we can add it separately with a narrower rationale.

PR #562, opened 2026-06-01
fix: render Chart with transparent background in both modes

AnmolSaini16:
ECharts' built-in "dark" theme paints the chart canvas with an opaque dark blue backgroundColor, which clashed with kumo's surface tokens (the surrounding bg-kumo-* should show through). Initialise with a symmetrical theme object in both modes and set backgroundColor to transparent.

Fixes #[insert GH or internal issue link(s)].

_Describe your change..._

---

- Reviews
- Tests
  - [ ] Tests included/updated


packages/kumo/src/components/chart/EChart.tsx
```diff
@@ -158,14 +158,23 @@ const transformTooltip = (tooltipObj: SafeTooltipOption) => {
   };
 };
 
-const prepareChartOptions = (options: KumoChartOption): EChartsOption => {
-  if (!options.tooltip) return options;
+const prepareChartOptions = (
+  options: KumoChartOption,
+  isDarkMode?: boolean,
```
mattrothenberg: Let's use a single object argument rather than positional arguments

mattrothenberg (changes requested):

packages/kumo/src/components/chart/EChart.tsx
AnmolSaini16: Fixed in latest commit.

mattrothenberg (approved):

PR #568, opened 2026-06-02
fix(tabs): improve scroll fades and active tab visibility

mattrothenberg:
## Summary

Fixes segmented `Tabs` scroll fades when the tab list only overflows by a few pixels, and keeps the selected tab visible when the tab list scrolls horizontally.

## Why

The previous scroll fade used CSS scroll-driven animations:

```css
animation-range:
  0 var(--scroll-fade-range, 3rem),
  calc(100% - var(--scroll-fade-range, 3rem)) 100%;
```

That works when the scroll range is larger than the default fade range (`3rem`). It breaks when tabs barely overflow, for example a `10px` scroll range with a `48px` fade range. In that case the left and right animation ranges overlap, so Chrome can compute a left fade even at `scrollLeft === 0`.

## What changed

- Segmented tabs now render `Tabs.List` as a Base UI `ScrollArea.Viewport`.
- The fade mask uses Base UI ScrollArea overflow distance variables:
  - `--scroll-area-overflow-x-start`
  - `--scroll-area-overflow-x-end`
- Fade width is clamped to the actual hidden content on each side, so tiny overflows render correctly.
- Selected tabs are scrolled into view when selected, with `scroll-padding-inline` matching the fade width so native scrolling keeps tabs out of the gradient zone.
- Underline tabs stay on the plain `Tabs.List` path to avoid clipping focus rings in a scroll viewport.
- Added a docs “Tiny Overflow” example to make this edge case easy to verify manually.
- Added focused unit coverage for the new segmented scroll/fade integration and selected-tab visibility behavior.

## Testing

- `pnpm --filter @cloudflare/kumo exec vitest run src/components/tabs/tabs.test.tsx`
- `pnpm --filter @cloudflare/kumo typecheck`
- `pnpm --filter @cloudflare/kumo lint` (passes with existing unrelated warnings)
- Built `@cloudflare/kumo` locally and manually verified in docs:
  - tiny-overflow segmented tabs render the correct edge fade
  - active segmented tab scrolls into view
  - underline tabs no longer clip focus rings
  - segmented tab list does not vertically scroll

- Reviews
- [x] automated review not possible because: visual scroll/fade behavior depends on browser layout and manual verification
- Tests
- [x] Tests included/updated
- [ ] Automated tests not possible - manual testing has been completed as follows: <description>
- [ ] Additional testing not necessary because: <your reason here>

### DIFF
--- a/.changeset/small-tabs-scroll.md
+++ b/.changeset/small-tabs-scroll.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Fix segmented Tabs scroll fades for tiny horizontal overflows and keep the selected tab visible when tab lists scroll.
--- a/packages/kumo-docs-astro/src/components/demos/TabsDemo.tsx
+++ b/packages/kumo-docs-astro/src/components/demos/TabsDemo.tsx
@@ -101,6 +101,42 @@ export function TabsManyDemo() {
   );
 }
 
+export function TabsTinyOverflowDemo() {
+  const [width, setWidth] = useState(596);
+
+  return (
+    <div className="w-full space-y-4">
+      <label className="flex max-w-md flex-col gap-2 text-sm text-kumo-subtle">
+        Container width: {width}px
+        <input
+          type="range"
+          min="520"
+          max="720"
+          value={width}
+          onChange={(event) => setWidth(Number(event.currentTarget.value))}
+        />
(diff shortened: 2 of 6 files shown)


mattrothenberg:
/bonk what do you think of this change?

mattrothenberg:
superseded by #569

PR #569, opened 2026-06-02
fix: proportional scroll fade

FBalint:
Fixes scroll-fade rendering, tab visibility on click, and ring placement for segmented `Tabs`.

### Changes

**Proportional scroll fade (CSS)**
The previous implementation used `mask-size` + `mask-composite: replace` to fade overflowing tab edges. This broke in some browsers because the multi-layer mask composition didn't interpolate proportionally during scroll-driven animations. Replaced it with two `@property`-registered custom properties (`--fade-left-n`, `--fade-right-n`) that animate from 0→1 and 1→0 respectively, driving a single `mask-image: linear-gradient(...)`. This is simpler and renders correctly across Chrome/Edge 115+.

**Scroll into view on click**
When tabs overflow, clicking a partially-visible tab now calls `scrollIntoView({ behavior: "smooth", block: "nearest", inline: "nearest" })` so the activated tab scrolls into the visible area. Also added `scroll-px-(--scroll-fade-width)` so the scroll snap accounts for the fade zone.

**Segmented ring placement**
Moved `ring ring-kumo-hairline/70` from the inner `TabsPrimitive.List` to the outer `TabsPrimitive.Root` so the border ring wraps the full segmented container instead of just the scrollable list.

- Reviews
  - [x] automated review not possible because: CSS scroll-driven animation behavior requires visual inspection
- Tests
  - [x] Automated tests not possible - manual testing has been completed as follows: verified scroll fade, scroll-into-view, and ring rendering in Chrome and Edge with overflowing segmented tabs in both light and dark mode

### DIFF
--- a/.changeset/fix-tabs-scroll-fade.md
+++ b/.changeset/fix-tabs-scroll-fade.md
@@ -1,0 +1,9 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Fix segmented `Tabs` scroll fade, scroll-into-view, and ring styling:
+
+- Rewrite CSS scroll-fade masking to use `@property`-animated custom properties, fixing proportional fade rendering across browsers.
+- Scroll the selected tab into view on click so it stays visible in overflowing tab lists.
+- Move `ring ring-kumo-hairline/70` from the inner list to the root container so the segmented variant ring wraps the entire component correctly.
--- a/packages/kumo/src/components/tabs/tabs.tsx
+++ b/packages/kumo/src/components/tabs/tabs.tsx
@@ -165,7 +165,8 @@ export function Tabs({
       {...rootProps}
       className={cn(
         "relative isolate min-w-0 font-medium",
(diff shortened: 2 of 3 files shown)


mattrothenberg:
We might have a small wrinkle with border radii (the corners look cut off)

Screenshot 2026-06-03 at 12 04 51 PM

FBalint:
> We might have a small wrinkle with border radii (the corners look cut off)
> 
> Screenshot 2026-06-03 at 12 04 51 PM

Fixed!

mattrothenberg:
Sorry for the nit-picky feedback. While we are here, can we fix this pre-existing focus ring border radius issue too?

Screenshot 2026-06-03 at 12 14 01 PM

FBalint:
> Sorry for the nit-picky feedback. While we are here, can we fix this pre-existing focus ring border radius issue too?

should be good now

mattrothenberg (approved):

PR #572, opened 2026-06-04
fix(sidebar): add text truncation with ellipsis to MenuButton content

hkfoster:
## Summary

`Sidebar.MenuButton` used `overflow-hidden` on its content span, which clipped long text but didn't show an ellipsis. Changed to `truncate` to match `Sidebar.MenuSubButton`'s existing behavior.

One-line change: `overflow-hidden` → `truncate` on the inner content `<span>`.

- Reviews
- [x] automated review not possible because: single-class CSS change

- Tests
- [x] Additional testing not necessary because: existing tests pass, no behavioral change — only adds `text-overflow: ellipsis` + `white-space: nowrap` to the existing `overflow: hidden`

### DIFF
--- a/.changeset/fix-menubutton-truncate.md
+++ b/.changeset/fix-menubutton-truncate.md
@@ -1,0 +1,7 @@
+---
+"@cloudflare/kumo": patch
+---
+
+fix(sidebar): add text truncation with ellipsis to Sidebar.MenuButton content
+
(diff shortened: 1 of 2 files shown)


mattrothenberg (approved):

mattrothenberg:
@hkfoster This works for me. One limitation here that may be intentional? This will **only** truncate primitive string/number children. If a consumer passes `<span>Very long label</span>` or a fragment containing text, this code leaves that element alone. That seems acceptable based on our current docs examples, but worth a callout. In those scenarios, you'd be on the hook for providing your truncation classes

PR #588, opened 2026-06-17
feat(chart): add enableLegendSelection prop to TimeseriesChart

AnmolSaini16:
### Summary
Adds an opt-in enableLegendSelection prop (default false) to TimeseriesChart. When enabled, it includes a hidden ECharts legend so consumers can drive series visibility imperatively via the legendSelect / legendUnSelect / legendToggleSelect actions — enabling custom interactive legends built with ChartLegend. Series toggled off via the legend are also excluded from the tooltip.

### Changes
- TimeseriesChart
- New enableLegendSelection?: boolean prop (default false); only adds legend: { show: false } when enabled.
- Tooltip now omits legend-deselected series, synced via legend events (legendselectchanged / legendselected / legendunselected) into a ref, avoids calling getOption() on the pointer-move hot path.
- Resets the selection ref when the prop is disabled so stale state can't filter the tooltip.

### Notes / Requirements
- Requires consumers to register ECharts' LegendComponent (echarts.use([LegendComponent])) when enableLegendSelection is on; otherwise the legend actions no-op (and dev logs a warning).
- Backward compatible: with the default false, no legend is added and tooltip behavior is unchanged.

---

- Reviews
- Tests
  - [ ] Tests included/updated
  - [x] Automated tests not possible - manual testing has been completed as follows: click through preview URL

### DIFF
--- a/.changeset/quick-legends-isolate.md
+++ b/.changeset/quick-legends-isolate.md
@@ -1,0 +1,12 @@
+---
+"@cloudflare/kumo": minor
+---
+
+feat(chart): add `enableLegendSelection` prop to `TimeseriesChart`
+
+Opt-in (default `false`) hidden ECharts legend that lets consumers drive series
+visibility imperatively via the `legendSelect` / `legendUnSelect` /
+`legendToggleSelect` actions — useful for building a custom interactive legend
+with `ChartLegend`. Series toggled off via the legend are also excluded from the
+tooltip. Requires registering ECharts' `LegendComponent`
+(`echarts.use([LegendComponent])`). When disabled, behaviour is unchanged.
--- a/packages/kumo-docs-astro/src/components/demos/Chart/ChartDemo.tsx
+++ b/packages/kumo-docs-astro/src/components/demos/Chart/ChartDemo.tsx
@@ -695,20 +695,25 @@ export function LegendOnClickDemo() {
     const chart = chartRef.current;
     if (!chart) return;
 
-    const isIsolated = series.every((s) =>
-      s.name === name ? !hiddenSeries[s.name] : hiddenSeries[s.name],
-    );
+    // Functional update so each click works from the latest state — rapid
+    // clicks won't race with React's state batching.
+    setHiddenSeries((prev) => {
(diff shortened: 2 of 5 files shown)


mattrothenberg:
Thanks for this! Clean opt-in API, nice perf approach with the ref instead of `getOption()`. Two questions:

1. Are `legendselected` and `legendunselected` needed (`TimeseriesChart.tsx:506-514`)? `legendselectchanged` already fires on any selection change and includes the full `selected` map. The other two typically only include `name`, so the `if (params?.selected)` guard skips them. Might be dead code?

2. Could the `params: any` on the legend handlers use the existing type instead? `ChartEvents` in `EChart.tsx:75-79` already types `legendselectchanged` as `{ name: string; selected: Record<string, boolean> }`. Dropping `any` should let TS infer it from `Partial<ChartEvents>`:
   ```ts
   legendselectchanged: (params) => {
     legendSelectedRef.current = params.selected;
   },
   ```

AnmolSaini16:
> Thanks for this! Clean opt-in API, nice perf approach with the ref instead of `getOption()`. Two questions:
> 
> 1. Are `legendselected` and `legendunselected` needed (`TimeseriesChart.tsx:506-514`)? `legendselectchanged` already fires on any selection change and includes the full `selected` map. The other two typically only include `name`, so the `if (params?.selected)` guard skips them. Might be dead code?
> 2. Could the `params: any` on the legend handlers use the existing type instead? `ChartEvents` in `EChart.tsx:75-79` already types `legendselectchanged` as `{ name: string; selected: Record<string, boolean> }`. Dropping `any` should let TS infer it from `Partial<ChartEvents>`:
>    ```ts
>    legendselectchanged: (params) => {
>      legendSelectedRef.current = params.selected;
>    },
>    ```

1.  They're required, `legendselectchanged` only fires for `legendToggleSelect` (and clicks on a real legend UI element), it does not fire for the` legendSelect/legendUnSelect` actions this feature uses, so listening to it alone misses every programmatic selection change.
2. Fixed the any type in latest commit.

mattrothenberg (approved):

PR #596, opened 2026-06-20
[kumo] Fix catalog schema initialization

pranavkatariain:
Fixes `initCatalog` so awaiting it loads the generated catalog schemas before synchronous validation methods are used.

Previously, `initCatalog` called `catalog.validateTree({})`, but validation only reads already-loaded schemas and does not trigger the async schema import. This meant the documented setup path could still return `Schemas not loaded`.

This change:
- calls `loadSchemas()` directly from `initCatalog`
- keeps the existing `initCatalog(catalog)` API shape for compatibility
- adds regression coverage for the documented setup path
- removes the stale `AGENTS.md` note

Note: `initCatalog` currently accepts a `catalog` argument even though schema loading is shared module state. This PR keeps that API shape for compatibility and only fixes the documented setup flow.

## Testing

- `pnpm --filter @cloudflare/kumo test src/catalog/catalog.test.ts --run`
- `pnpm --filter @cloudflare/kumo typecheck`
- `git diff --check main...HEAD`

- Reviews
  - [x] automated review not possible because: small catalog initialization fix reviewed manually
- Tests
  - [x] Tests included/updated
  - [ ] Automated tests not possible - manual testing has been completed as follows: n/a
  - [ ] Additional testing not necessary because: n/a


stritt:
thanks for this, the fix is nice and focused and still looks useful. it mainly needs an update from current main so we can get fresh ci results, then it should be ready for review.

pranavkatariain:
@stritt Thanks! I have updated the branch with current main.

TheAlexLichter:
Oh whoops, I did not see this PR and fixed it in #688

pranavkatariain:
All good! I will close this PR. Thanks!

PR #601, opened 2026-06-24
fix(dialog): use fixed width instead of min-width for size prop

lkipke:
## Summary

The Dialog `size` prop only controlled `min-width`, so content wider than the minimum would stretch the dialog. This changes the size variants to set a fixed `width` on desktop, making the `size` prop behave as expected.

### Root cause

- Base styles included `sm:w-auto` which made the dialog shrink-to-fit content
- Size variants used `min-w-*` classes (floor only, no ceiling)
- Any wide content (tables, long text) pushed the dialog past its intended size

### Fix

- Removed `sm:w-auto` from base dialog styles
- Changed size variants from `min-w-*` to `sm:w-*` (fixed width on desktop)
- Mobile stays `w-full` via the `sm:` breakpoint prefix
- `max-w-[calc(100vw-2rem)]` still prevents viewport overflow

| Size | Before | After |
|------|--------|-------|
| `sm` | `min-w-72` | `sm:w-72` (288px) |
| `base` | `sm:min-w-96` | `sm:w-96` (384px) |
| `lg` | `min-w-[32rem]` | `sm:w-[32rem]` (512px) |
| `xl` | `min-w-[48rem]` | `sm:w-[48rem]` (768px) |

### Demo

Added a "Sizes" demo to the docs site showing all four sizes with a wide table inside each dialog, so you can verify the dialog holds its width.

- Reviews
- [x] automated review not possible because: visual/CSS change requires manual inspection
- Tests
- [x] Additional testing not necessary because: CSS-only change with docs demo for manual verification

### DIFF
--- a/.changeset/fix-dialog-size-width.md
+++ b/.changeset/fix-dialog-size-width.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Fix Dialog `size` prop to set a fixed width instead of only a minimum width. Previously, dialog content could stretch the dialog beyond its intended size.
--- a/packages/kumo-docs-astro/src/components/demos/DialogDemo.tsx
+++ b/packages/kumo-docs-astro/src/components/demos/DialogDemo.tsx
@@ -6,6 +6,7 @@ import {
   Combobox,
   DropdownMenu,
 } from "@cloudflare/kumo";
+import type { DialogProps } from "@cloudflare/kumo";
 import { Warning, X } from "@phosphor-icons/react";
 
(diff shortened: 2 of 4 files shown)


.changeset/fix-dialog-size-width.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
```
mattrothenberg, line 2: Only need one changeset

mattrothenberg (changes requested):

mattrothenberg (approved):

PR #570, opened 2026-06-03
feat: add LayerDialog component

invisal:
This PR introduces the `LayerDialog` component, a modal dialog built on top of `@base-ui/react/dialog` and styled with `LayerCard`.

The component includes:

- `LayerDialog.Root` — manages the dialog's open/close state
- `LayerDialog.Trigger` — opens the dialog
- `LayerDialog.Close` — closes the dialog
- `LayerDialog.Content` — renders the dialog inside a portal with a backdrop, centered on screen
- `LayerDialog.Body` — primary content area
- `LayerDialog.Footer` — action area at the bottom
- `LayerDialog.Title` — dialog heading
- `LayerDialog.Description` — descriptive text
- `LayerDialog.Separator` — horizontal divider between sections

Also includes documentation page, interactive demo, package export, and a changeset.

image

## Examples

```tsx
<LayerDialog>
  <LayerDialog.Trigger render={<Button>Open</Button>} />
  <LayerDialog.Content>
     <LayerDialog.Body>
        <LayerDialog.Title />
        <LayerDialog.Description />
     </LayerDialog.Body>
     <LayerDialog.Footer className="justify-between">
        <LayerDialog.Close render={<Button>Close</Button>} />
        <Button>Save</Button>
     </LayerDialog.Footer>
  </LayerDialog.Content>
</LayerDialog>
```

### Reviews
- [ ] bonk has reviewed the change
- [x] automated review not possible because: new component requires design and API review

PR #603, opened 2026-06-24
feat(dialog): add layer variant

geoquant:
## Summary

This PR provides an alternative to #570 by bringing the Layer Dialog treatment into the existing `Dialog` API instead of introducing a new top-level `LayerDialog` component.

- Adds `variant="layer"` to `Dialog`, with the existing behavior preserved as the default `variant="base"`
- Adds additive structured subcomponents: `Dialog.Body`, `Dialog.Footer`, and `Dialog.Separator`
- Keeps `Dialog.Title` and `Dialog.Description` backed by the Base UI dialog primitives for accessibility semantics
- Documents both variants in the Dialog docs and adds layer variant demos
- Incorporates Nanda's review notes from #570 by using a larger outer radius and equalized layer padding

## Validation

- `pnpm --filter @cloudflare/kumo lint` — passes with existing warnings
- `pnpm --filter @cloudflare/kumo typecheck` — passes
- `cd packages/kumo && pnpm exec vitest run --project=unit src/components/dialog/dialog.test.tsx` — passes
- `pnpm --filter @cloudflare/kumo codegen:registry` — passes; generated outputs are ignored/untracked

Note: `pnpm --filter @cloudflare/kumo-docs-astro typecheck` currently fails on unrelated existing docs/demo type errors across SkeletonLine, Switch, Table, Toolbar, Tooltip, Chart, SankeyChart, and TableOfContents.

- Reviews
- [x] automated review not possible because: this is an API/design-system direction change that needs human review against #570
- Tests
- [x] Tests included/updated
- [ ] Automated tests not possible - manual testing has been completed as follows: n/a
- [ ] Additional testing not necessary because: n/a


geoquant:
I reviewed the Dialog usages scraped from Stratus with a focus on backwards compatibility for existing className/style extension patterns. I did not find a breaking change for existing/default Dialog usage.

Compatibility confirmed:
- Existing `<Dialog className=...>` / `<Dialog.Content className=...>` surface overrides continue to merge onto the dialog surface.
- `Dialog.Title`, `Dialog.Description`, `Dialog.Trigger`, and `Dialog.Close` continue to accept/forward className/props.
- I added coverage for common production patterns like `p-0`, `max-w-lg`, `rounded-none`, custom background classes, important title typography overrides, dark-mode description classes, and `data-testid` passthrough.
- I also fixed prop passthrough on Dialog content so arbitrary popup props like `data-testid`, `id`, and aria attributes remain supported.

One difference to document, but not a breaking change: for the new `variant="layer"`, `Dialog.Content className` applies to the outer popup wrapper. That supports documented usage such as changing width with `className="max-w-lg"`. The inner visible `LayerCard` still owns its own surface classes such as background, padding, rounded corners, and shadow. Since `variant="layer"` is new, this does not break existing users, but it is worth being explicit about in docs if people expect layer dialogs to restyle the inner card through `Dialog.Content className`.

PR #605, opened 2026-06-25
feat(chart): add hideXAxisLabels and hideYAxisLabels props to TimeseriesChart

lkipke:
Adds two new boolean props to `TimeseriesChart` that allow consumers to hide axis tick labels:

- `hideXAxisLabels` — hides x-axis timestamp labels
- `hideYAxisLabels` — hides y-axis value labels

Both default to `false` (labels visible). Axis lines and grid lines are unaffected.

### Usage

```tsx
<TimeseriesChart
  echarts={echarts}
  data={data}
  hideXAxisLabels
  hideYAxisLabels
/>
```

Passes through to ECharts' `axisLabel.show` on the respective axis config.

- Reviews
  - [x] automated review not possible because: trivial prop passthrough to ECharts
- Tests
  - [x] Additional testing not necessary because: passthrough boolean props with no logic beyond negation

### DIFF
--- a/.changeset/hide-axis-labels.md
+++ b/.changeset/hide-axis-labels.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": minor
+---
+
+Add `hideXAxisLabels` and `hideYAxisLabels` props to TimeseriesChart for hiding axis tick labels.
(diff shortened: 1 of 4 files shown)


lkipke:
Closing as the appetite is to create a new Sparkline component instead

PR #622, opened 2026-07-09
feat: add timeseries thresholds

Ankcorn:
Adds a first-class `thresholds` API to `TimeseriesChart` for horizontal value-axis reference lines, e.g. memory limits or quota boundaries. Thresholds require a value and color; labels are optional for cases where an unlabeled reference line is useful.

This keeps threshold rendering inside the high-level chart abstraction and uses ECharts `markLine` through a small dedicated helper, matching the marker implementation without adding tooltip-specific behaviour. Threshold values also participate in y-axis extent calculation so limits above/below the current data remain visible.

image

- Reviews
- [x] automated review not possible because: visual chart annotation behaviour needs human review
- Tests
- [ ] Tests included/updated
- [ ] Automated tests not possible - manual testing has been completed as follows: checked docs demo locally
- [x] Additional testing not necessary because: existing chart behaviour is unchanged when no thresholds are provided

### DIFF
--- a/.changeset/timeseries-thresholds.md
+++ b/.changeset/timeseries-thresholds.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": minor
+---
+
+Add horizontal threshold lines to TimeseriesChart.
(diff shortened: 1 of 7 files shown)


ibbykhazanchi (approved):
LGTM

mattrothenberg (approved):

PR #628, opened 2026-07-10
feat(banner): adjust banner colors and add size + actions props

najlaskr:
Fixes muddy CTA colors that washed out against banner backgrounds, adjust banner hues for better contrast and adds new size options.

- Modified existing tokens to better align with banner backgrounds, text and CTA combinations that reflect the different variants
- Added a new `Banner.Actions` CTA compound to allow specific-styling for action buttons without affecting existing `Button` styles. `Banner.Action` supports `primary` (filled), `secondary` (accent-hued outline on a transparent background), and `ghost` variants
- Added a `size` prop to `Banner` (`"base"` | `"sm"`); the compact `"sm"` size suits dialogs and other tight spaces and defaults its `Banner.Action` children to the `xs` size

### Before
CleanShot 2026-07-10 at 16 12 45@2x
CleanShot 2026-07-10 at 16 14 24@2x

### After
CleanShot 2026-07-10 at 16 12 58@2x
CleanShot 2026-07-10 at 16 13 44@2x

---

- Reviews
  - [x] automated review not possible because: no access to bonk; mostly visual changes to banner
- Tests
  - [ ] Tests included/updated
  - [x] Automated tests not possible - manual testing has been completed as follows: visual testing on banners, ran `pnpm lint`, `pnpm typecheck` and `pnpm test`


mattrothenberg (changes requested):
Love the visuals here, the after screenshots are a major improvement! A few things before merging:

1. I think Banner.Actions is the wrong API: it's just a flex div, and since it has exactly one correct usage it adds boilerplate for every consumer. The Banner root could probably apply that layout to the action slot itself.
2. Relatedly, Banner.Action is a parallel fork of Button (its own size table, variant names, and focus styles, and the sizes already diverge from KUMO_BUTTON_VARIANTS despite the comment saying they mirror it). Open question: is Button's forced !text-white the only blocker to reusing it? If so I'd rather fix that in Button than maintain a second button that will drift.
3. Third, the theme changes rewrite global tokens (--color-kumo-warning, the *-tint family, etc.) that toast, badge, and command palette also consume, so this changes more than banners while testing only covers banners. Was this change intentional? Am I misreading the diff?

stritt:
thanks for this! we'd still like to see if we can get it merged. the main things to work through are simplifying the banner action api and either keeping the token changes scoped to banner or adding coverage for the other components they affect. once those are updated we can get it reviewed again.

najlaskr:
> Love the visuals here, the after screenshots are a major improvement! A few things before merging:
> 
> 1. I think Banner.Actions is the wrong API: it's just a flex div, and since it has exactly one correct usage it adds boilerplate for every consumer. The Banner root could probably apply that layout to the action slot itself.
> 2. Relatedly, Banner.Action is a parallel fork of Button (its own size table, variant names, and focus styles, and the sizes already diverge from KUMO_BUTTON_VARIANTS despite the comment saying they mirror it). Open question: is Button's forced !text-white the only blocker to reusing it? If so I'd rather fix that in Button than maintain a second button that will drift.
> 3. Third, the theme changes rewrite global tokens (--color-kumo-warning, the *-tint family, etc.) that toast, badge, and command palette also consume, so this changes more than banners while testing only covers banners. Was this change intentional? Am I misreading the diff?

updated these in the new commits!

mattrothenberg (approved):

PR #641, opened 2026-07-17
Prevent Generated Chart Search Results from Pointing to Nonexistent Docs

anott03:
Currently the auto-generated search results for Chart components point to nonexistent paths: `/components/chart`, `/components/bubble-map`, etc. Most of these have guides already, so the generated search results were redundant, so removing them was sufficient. `BubbleMap` and `ChoroplethMap`, however, are documented in the "Maps" guide, which is not listed as a result when searching for those components. In those cases, add overrides that direct them to `/charts/maps`.

---

- Reviews
- Tests
  - [ ] Tests included/updated


ArshVermaGit (commented):
Nice cleanup. Removing the invalid doc links and redirecting the map components to the correct documentation improves the search experience. LGTM.

PR #644, opened 2026-07-20
[Docs] Fix broken chart links in search

PedroLunet:
Fixes the docs command-palette search linking chart components to non-existent pages.

Searching for a chart (e.g. "chart", "bubble", "sankey") returned results linking to `/components/<name>` (e.g. `/components/bubble-map`) — all 404s. Chart components are auto-discovered from the component registry, and `getComponentUrl()` builds their URLs as `/components/<name>`, but chart docs actually live under `/charts/*`.

## Fix

- **Link charts to their real pages.** A new `CHART_COMPONENT_URLS` map sends `SankeyChart` → `/charts/sankey`, `TimeseriesChart` → `/charts/timeseries`, and both map components to `/charts/maps`. `getComponentUrl()` checks this map first, so these stay searchable and land on the right page instead of 404ing.
- **Deep-link the shared page.** `BubbleMap` and `ChoroplethMap` share `/charts/maps`, so they link to their section anchors (`#bubble-map`, `#choropleth-map`), matching `Heading.astro`'s slug generation.
- **Exclude only pages that don't exist standalone.** `Chart`/`ChartLegend` have no dedicated page (already covered by the curated "Charts" and "Custom Chart" `STATIC_PAGES` entries), so they're filtered via `CHART_COMPONENTS_WITHOUT_OWN_PAGE`.
- Removed the now-redundant hand-written "Sankey Chart" / "Timeseries Chart" static entries (the registry components link directly now); kept "Maps" since that page hosts two components.

Net result: every chart stays searchable (better than hiding them) and links to correct content.

---

- Reviews
  - [x] automated review not possible because: bonk workflow was cancelled on this fork PR; reviewed locally with Cloudflare's ai-review CLI (approved, 0 issues)
- Tests
  - [ ] Tests included/updated
  - [x] Additional testing not necessary because: client-side search URL mapping only; verified with `astro check` (0 errors) and oxlint (0 errors), and confirmed the target `/charts/*` pages and `#bubble-map` / `#choropleth-map` heading anchors exist


mattrothenberg:
@PedroLunet am I doing something silly here?
Screenshot 2026-07-20 at 1 20 39 PM

PedroLunet:
> @PedroLunet am I doing something silly here? Screenshot 2026-07-20 at 1 20 39 PM

@mattrothenberg I'm investigating it further, but that BubbleMap component doesn't seem to have a page (at least not one that I can find). In prod if you search for it and click the option on the search bar, it lead to a page that doesn't exist. (https://kumo-ui.com/components/bubble-map), the issue i tried to fix. I'll take a deeper look and share my findings!

PR #653, opened 2026-07-23
fix(chart): constrain timeseries tooltip to chart

AnmolSaini16:
Fixes the TimeseriesChart tooltip following the cursor outside the chart after a native context-menu interaction.

Adds chart-boundary detection while the tooltip is open, providing a fallback when ECharts does not emit `globalout`. Includes regression coverage for the affected interaction.

**Reviews**

**Tests**
- [x] Tests included/updated


mattrothenberg:
/bonk do we need an SSR safe way of doing this?

PR #658, opened 2026-07-23
build: migrate toolchain to Vite+ (vp)

TheAlexLichter:
This is the first of a few PRs to move the repo over to Vite+.

What that means:

* vp test/vp lint/vp fmt replacing vitest/oxlint/prettier
* THe pnpm catalog aliases vite/vitest to the Vite+ packages
* The three `.oxlintrc.json` files are merged into `lint` blocks in `vite.config.ts`
  * Custom rules now import  `defineRule` from `@oxlint/plugins`
  * Native jsx-a11y rules are used instead of the JS plugin
*  Lefthook is replaced with vp's git hook management
* Rewrites 91 `vitest` imports to `vite-plus/test`
* Add temporary Prettier-parity `fmt.ignorePatterns` with upstream-issue links for ~10 files.
* Make `use-sync-external-store` a real dependency (more accurate than bundling it)
* Kumo is now built with Vite 8/Rolldown (this is the biggest "danger" for the user ⚠️)
  * I did test the build but would appreciate a double check 	
---

- Reviews
- Tests
  - [ ] Tests included/updated
  - [x] Automated tests not possible - manual testing has been completed as follows: Running commands, checking outputs and testing built kumo
  - [x] Additional testing not necessary because: Matt did a thorough manual review.


packages/kumo-docs-astro/src/env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="astro/client" />
```
TheAlexLichter, line 1: Necessary so vp check understands the types used (usually auto generated but that is "too late")

packages/kumo-docs-astro/tsconfig.json
```diff
@@ -3,9 +3,8 @@
   "include": [".astro/types.d.ts", "**/*"],
   "exclude": ["dist"],
   "compilerOptions": {
-    "baseUrl": ".",
     "paths": {
-      "~/*": ["src/*"]
+      "~/*": ["./src/*"]
```
TheAlexLichter, line 7: Needed for TS7

packages/kumo-figma/tsconfig.json
```diff
@@ -10,7 +10,6 @@
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
     "types": ["@figma/plugin-typings", "node"],
-    "baseUrl": ".",
```
TheAlexLichter, line 13: Also needed for TS7 compat.

.node-version
```diff
@@ -0,0 +1 @@
+24.12.0
```
TheAlexLichter, line 1: Added in addition to .nvmrc so previous tooling still works. This might cause drift in the future though.

pnpm-workspace.yaml
```diff
@@ -31,10 +31,33 @@ catalog:
...
+  vite: "catalog:"
+  vitest: "catalog:"
+peerDependencyRules:
+  allowAny:
+    - vite
+    - vitest
+  allowedVersions:
+    vite: "*"
+    vitest: "*"
+minimumReleaseAgeExclude:
```
TheAlexLichter, line 53: We can remove these exceptions too.

packages/kumo/tests/imports/import-entries.mjs
TheAlexLichter, line 1: Added as "smoke test" to avoid interop issues

mattrothenberg (approved):

PR #671, opened 2026-07-27
feat(clipboardText): spam-safe copy feedback and toast bump animation

nocdn:
### Summary of changes

Improves `ClipboardText` copy feedback when the button is pressed repeatedly:

- The check icon stays 1.5s after the **last** click instead of after the first, fixing 
a "glitching" of the copy/check icons fighting for placement (essentially timeout is cleared and restarted on each press, so it no longer flickers back to the copy icon while spamming)

| Before | After |
|--------|--------|
|spambefore | spamafter |

- Only one "Copied" toast for the ClipboardText with tooltip, instead of a stack of toasts. Also includes a subtle "bump" animation to the one tooltip when copying again, instead of adding another one.

| Before | After |
|--------|--------|
| tooltipspambefore | tooltipspamafter |

---

- Reviews
- Tests
  - [x] Tests included/updated

### DIFF
--- a/.changeset/clipboard-text-copy-feedback.md
+++ b/.changeset/clipboard-text-copy-feedback.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+ClipboardText: keep the check icon until the last copy click settles, and bump the anchored "Copied" toast (without stacking) when the button is pressed again while it's open.
--- a/packages/kumo/src/components/clipboard-text/clipboard-text.test.tsx
+++ b/packages/kumo/src/components/clipboard-text/clipboard-text.test.tsx
@@ -1,0 +1,150 @@
+import { act, fireEvent, render, screen } from "@testing-library/react";
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  it,
+  vi,
+} from "vite-plus/test";
+import { ClipboardText } from "./clipboard-text";
+
+describe("ClipboardText", () => {
+  let writeText: ReturnType<typeof vi.fn>;
+
+  beforeEach(() => {
+    vi.useFakeTimers();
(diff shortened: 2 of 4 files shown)


mattrothenberg (changes requested):
Requesting changes on the approach, not the feature.

The behavior sounds right: repeated copy clicks should not stack toasts, the copied state should last until the final click settles, and the visible toast should react when clicked again.

I think this can be much simpler. `ClipboardText` now uses a shared module-level toast manager, but the toast ID comes from `useId()`. `useId()` is only unique within a React root. In multi-root cases like Astro islands or embedded widgets, two `ClipboardText` instances can get the same ID and update or close each other’s toast.

Base UI appears to already support the behavior we need here. Adding a toast with an existing ID updates it in place and refreshes its timeout, so we should not need to separately track `toastOpenRef` and branch between `add` and `update`. Instead, we can store the ID returned from `clipboardToastManager.add()` in a ref and pass it back on the next click.

Roughly:

```tsx
const toastIdRef = useRef<string | null>(null);

toastIdRef.current = clipboardToastManager.add({
  id: toastIdRef.current ?? undefined,
  description: copiedText,
  positionerProps: {
    anchor: buttonRef.current,
    side: tooltipSide,
    sideOffset: 8,
  },
  timeout: COPIED_FEEDBACK_MS,
  onClose() {
    toastIdRef.current = null;
    setCopied(false);
  },
});
```

For the bump animation, we can likely key off Base UI’s `toast.updateKey` rather than carrying custom `data.bumpKey`.

Can we simplify this around Base UI’s existing upsert behavior and avoid the `useId()` collision risk?

nocdn:
Thanks for the feedback, let me look into it and I'll make the changes here

mattrothenberg (approved):

PR #674, opened 2026-07-27
fix(banner): quieten and brighten the Banner.Action CTA

AnmolSaini16:
`Banner.Action` defaulted to `variant="primary"`, so every banner rendered a saturated filled accent button that overpowered its own message and competed with page-level CTAs.

- Default to `variant="secondary"`: an accent-tinted fill with an accent hairline. `variant="primary"` stays available for the rare banner whose CTA is the most important action on the page.
- Brighten the `secondary` and `ghost` accents. Labels and icons now use the banner's accent text token directly instead of inheriting the container colour, which picked up the neutral banner's dimmed `text-kumo-default/70` and made the CTA look muddy.
- Gate hover colours behind `not-disabled:`, so hover is inert on disabled and loading actions. This also drops the `!important` hovers, since the modifier set now matches Button's own outline defaults and tailwind-merge dedupes them.

---

- Reviews
- Tests
  - [ ] Tests included/updated


ArshVermaGit (commented):
Looks good to me. The updated CTA styling feels more balanced, and the color and hover state adjustments improve clarity while keeping the component consistent with the design system. The changes are clean and well scoped. LGTM.

PR #711, opened 2026-08-06
fix: forward link props in Tabs and Sidebar

### LINKED ISSUE #690: Tabs: `nativeButton` is not forwarded to Base UI's `Tabs.Tab`, so link-based tabs always warn
## Summary

`Tabs` documents `TabsItem.render` as being for link-based tabs, but does not forward
Base UI's `nativeButton` prop to `Tabs.Tab`. Since Base UI defaults `nativeButton` to
`true`, every link-based tab logs a dev warning on mount, and there is no way to
silence it from application code.

> A component that acts as a button expected a native `<button>` because the
> `nativeButton` prop is true. Rendering a non-`<button>` removes native button
> semantics, which can impact forms and accessibility. Use a real `<button>` in the
> `render` prop, or set `nativeButton` to `false`.

The warning is dev-only and I could not find any user-visible breakage (see
*Impact* below), but it fires once per tab bar per mount, so it shows up on every mount in apps
that follow the documented link-tab pattern.

## Reproduction

```tsx
import { Link } from "@cloudflare/kumo/components/link";
import { Tabs } from "@cloudflare/kumo/components/tabs";

<Tabs
  variant="underline"
  value={current}
  tabs={[
    { value: "overview", label: "Overview",
      render: <Link href="/settings/overview" className="no-underline" /> },
    { value: "members", label: "Members",
      render: <Link href="/settings/members" className="no-underline" /> },
  ]}
/>
```

**Expected:** tabs render as real `<a href>` (so middle-click / copy-link-address /
context menu work) with no console warning.

**Actual:** tabs render correctly and navigate correctly, but the warning above is
logged on every mount. Measured with a Playwright probe (`page.on("console")`): one
warning per page that has a link-based tab bar, zero on pages without one.

## Root cause

`Tabs` hand-picks the props it passes to Base UI's `Tabs.Tab`, and there is no rest
spread on either `TabsProps` or `TabsItem` — so `nativeButton` cannot be reached from
application code at all:

```js
// dist/chunks/tabs-*.js (2.9.0, unminified)
items.map((tab) => jsx(TabsTab, {
  "data-kumo-component": "Tabs",
  "data-kumo-part": "tab",
  value: tab.value,
  render: tab.render,
  onClick: (e) => { e.currentTarget.scrollIntoView({ /* … */ }); },
  className: cn(/* … */, tab.className),
  children: tab.label
}, tab.value))
```

Base UI itself supports the prop — it is only the pass-through that is missing:

```js
// @base-ui/react 1.6.0 — tabs/tab/TabsTab
const { className, disabled = false, render, value, id: idProp,
        nativeButton = true, style, ...elementProps } = componentProps;
// …
const { getButtonProps, buttonRef } = useButton({
  disabled, native: nativeButton, focusableWhenDisabled: true,
});
```

`TabsItem` documents this use case, which is what sent us looking for the prop in the
first place:

```ts
// dist/src/components/tabs/tabs.d.ts
/**
 * Custom render function or element to replace the tab element (e.g. for link-based tabs).
 * …
 */
render?: TabsTab.Props["render"];
```

**Affected versions:** reproduced on **2.8.0** and **2.9.0** (latest at time of
writing). The tab-construction code is identical in both, so upgrading does not help.

## Impact

Low, and I want to be accurate about it rather than overstate:

`nativeButton` feeds Base UI's `useButton`, where it gates `tabIndex`, `aria-disabled`
and the `disabled` attribute. But `TabsItem` has no `disabled` field and `Tabs` never
passes `disabled` to `Tabs.Tab`, so those branches can't produce a user-visible
difference through Kumo's public API. In practice the cost is the console noise —
which is still worth fixing, because the pattern that triggers it is the one the
docs recommend.

## Proposed fix

Forward the prop from the item descriptor:

```ts
export type TabsItem = {
  value: string;
  label: ReactNode;
  className?: string;
  render?: TabsTab.Props["render"];
  /** Set to false when `render` produces a non-<button> element (e.g. a link). */
  nativeButton?: boolean;
};
```

```js
jsx(TabsTab, {
  // …
  nativeButton: tab.nativeButton,
})
```

I'd suggest an explicit prop rather than inferring it from `render` — `render` can
legitimately be a real `<button>`, so a heuristic would guess wrong in that case.

## Workaround

For anyone hitting this before it's fixed: dropping to the Base UI primitives that
Kumo already re-exports works, at the cost of restating the tab styling.

```tsx
import { Tabs as TabsPrimitive } from "@cloudflare/kumo/primitives/tabs";

<TabsPrimitive.Tab value={item.value} nativeButton={false} render={<Link href={item.href} />}>
  {item.label}
</TabsPrimitive.Tab>
```

## Related

#367 is the same Base UI warning surfacing through a different component
(`DropdownMenu.Trigger`).

## Environment

- `@cloudflare/kumo` 2.8.0 and 2.9.0
- `@base-ui/react` 1.6.0 (as bundled)
- React 19.2, Chromium via Playwright 1.62

mattrothenberg:
## Summary
- add `nativeButton` to `TabsItem` and forward it to Base UI `Tabs.Tab` so link-rendered tabs can opt out of native button semantics
- preserve extra props in Sidebar link branches so attributes like `aria-current` reach `MenuButton` and `MenuSubButton` anchors
- update the Tabs render-prop demo/docs for link-rendered tabs
- add regression tests for Tabs link rendering and Sidebar link attribute forwarding

Closes #690

## Tests
- pnpm --filter @cloudflare/kumo test:unit src/components/tabs/tabs.test.tsx src/components/sidebar/sidebar.test.tsx
- pnpm --filter @cloudflare/kumo lint
- pnpm format:check

- Reviews
- [x] automated review not possible because: targeted component API fixes with regression tests
- Tests
- [x] Tests included/updated
- [ ] Additional testing not necessary because: not applicable

### DIFF
--- a/.changeset/clean-link-prop-forwarding.md
+++ b/.changeset/clean-link-prop-forwarding.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Forward link-specific props in Tabs and Sidebar so link-rendered tabs can opt out of native button semantics, and Sidebar link buttons preserve attributes like `aria-current`.
--- a/.changeset/gentle-tabs-link-render.md
+++ b/.changeset/gentle-tabs-link-render.md
@@ -2,4 +2,4 @@
 "@cloudflare/kumo": patch
 ---
 
-Forward `nativeButton` from Tabs items to support link-rendered tabs without Base UI warnings.
+Forward link-specific props in Tabs and Sidebar link rendering paths.
--- a/.changeset/gentle-tabs-link-render.md
+++ b/.changeset/gentle-tabs-link-render.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/kumo": patch
+---
+
+Forward `nativeButton` from Tabs items to support link-rendered tabs without Base UI warnings.
--- a/.changeset/gentle-tabs-link-render.md
+++ b/.changeset/gentle-tabs-link-render.md
@@ -1,5 +1,0 @@
----
-"@cloudflare/kumo": patch
----
-
-Forward link-specific props in Tabs and Sidebar link rendering paths.
--- a/packages/kumo-docs-astro/src/components/demos/TabsDemo.tsx
+++ b/packages/kumo-docs-astro/src/components/demos/TabsDemo.tsx
@@ -165,11 +165,13 @@ export function TabsRenderPropDemo() {
         {
           value: "tab2",
           label: "Link Tab",
+          nativeButton: false,
           render: (props) => <a {...props} href="#tab2" />,
         },
         {
           value: "tab3",
           label: "Cloudflare",
+          nativeButton: false,
           render: (props) => (
             <a {...props} href="https://cloudflare.com" target="_blank" />
           ),
--- a/packages/kumo-docs-astro/src/pages/components/tabs.mdx
+++ b/packages/kumo-docs-astro/src/pages/components/tabs.mdx
@@ -167,6 +167,16 @@ export default function Example() {
         <td class="px-4 py-3 font-mono text-xs">string</td>
         <td class="px-4 py-3 font-mono text-xs">No</td>
       </tr>
+      <tr class="border-b border-kumo-hairline">
+        <td class="px-4 py-3 font-mono text-xs">render</td>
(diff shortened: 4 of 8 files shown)


mattrothenberg:
/bonk you keep saying approved but haven't actually approved

PR #726, opened 2026-08-11
fix: align banner and toast icons

mattrothenberg:
## Summary

- Align Banner icons to the first line of wrapping message text using a line-height-sized icon slot
- Prevent Banner and Toast status icons from shrinking when messages wrap
- Add a Banner docs before/after preview with copyable Banner API code

Screenshot 2026-08-11 at 9 54 45 AM

## Tests

- pnpm --filter @cloudflare/kumo exec vp test run --project=unit src/components/banner/banner.test.tsx
- pnpm --filter @cloudflare/kumo lint
- pnpm --filter @cloudflare/kumo typecheck
- pnpm --filter @cloudflare/kumo-docs-astro exec vp lint src/components/demos/BannerDemo.tsx src/pages/components/banner.mdx
- pnpm --filter @cloudflare/kumo-docs-astro codegen:demos

## Notes

Full docs lint currently fails on unrelated existing issues in TabsDemo.tsx, ToolbarDemo.tsx, and ChartCard.astro.

- Reviews
- [x] automated review not possible because: this is a focused visual alignment fix that needs human visual review
- Tests
- [x] Tests included/updated

### DIFF
--- a/packages/kumo/src/components/banner/banner.test.tsx
+++ b/packages/kumo/src/components/banner/banner.test.tsx
@@ -156,7 +156,7 @@ describe("Banner", () => {
 
     const icon = screen.getByTestId("icon");
     const iconClassName = icon.getAttribute("class") ?? "";
-    expect(iconClassName).toContain("h-[1lh]");
+    expect(iconClassName).toContain("size-[1em]");
     expect(iconClassName).toContain("flex-none");
     expect(iconClassName).toContain("custom-icon");
     expect(icon.parentElement?.className).toContain("h-[1lh]");
--- a/packages/kumo/src/components/banner/banner.test.tsx
+++ b/packages/kumo/src/components/banner/banner.test.tsx
@@ -139,11 +139,28 @@ describe("Banner", () => {
     expect(className).toContain("px-3");
     expect(className).toContain("py-2");
     expect(className).toContain("text-sm");
-    // Compact banners align everything on one centered row.
-    expect(className).toContain("items-center");
+    // Compact banners align icons to the first text line when content wraps.
+    expect(className).toContain("items-start");
     // Base-size spacing/alignment must not leak in.
     expect(className).not.toContain("px-4");
-    expect(className).not.toContain("items-start");
+  });
+
+  it("aligns icons to the first text line without shrinking", () => {
+    render(
+      <Banner
+        size="sm"
+        icon={<svg data-testid="icon" className="custom-icon" />}
+        description="A DNS record already exists in this zone and may wrap onto multiple lines."
+      />,
+    );
+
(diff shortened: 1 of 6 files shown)


packages/kumo/src/components/banner/banner.test.tsx
```diff
@@ -139,11 +139,28 @@ describe("Banner", () => {
...
+      />,
+    );
+
+    const icon = screen.getByTestId("icon");
+    const iconClassName = icon.getAttribute("class") ?? "";
+    expect(iconClassName).toContain("size-[1em]");
+    expect(iconClassName).toContain("flex-none");
+    expect(iconClassName).toContain("custom-icon");
+    expect(icon.parentElement?.className).toContain("h-[1lh]");
+    expect(icon.parentElement?.className).toContain("flex-none");
```
nandanmen, line 163: **nit**: we shouldn't test on the existence of class names, it's brittle

mattrothenberg: truth! I will fix this

packages/kumo/src/components/banner/banner.tsx
```diff
@@ -63,28 +65,51 @@ export type KumoBannerSize = keyof typeof KUMO_BANNER_VARIANTS.size;
...
 
+const renderBannerIcon = (icon: ReactNode, className?: string) => {
+  const iconElement = isValidElement(icon)
+    ? (icon as ReactElement<{ className?: string }>)
+    : null;
+  const alignedIcon = iconElement
+    ? cloneElement(iconElement, {
+        className: cn(
+          iconElement.props.className,
+          "size-[1em] flex-none leading-snug",
```
geoquant, line 96: I checked downstream Stratus usage and this does have compatibility impact: there are at least 123 `Banner` usages across 90 files that pass explicit numeric icon sizes, including many 18px, 20px, and 24px icons. Appending `size-[1em]` overrides those sizes for Phosphor icons (for example, a requested 20px icon becomes the Banner’s inherited `1em` size), while custom icon components that do not forward `className` may retain their original size, so the normalization is inconsistent across the current `ReactNode` contract.

Is taking ownership of all Banner icon sizing intentional here? If the goal is only alignment/non-shrinking, the outer `h-[1lh] flex-none items-center` wrapper appears sufficient without cloning or rewriting the supplied node. If normalization is intentional, I think it should be called out as an API/visual behavior change because Stratus currently relies heavily on explicit icon sizing.

mattrothenberg: Intentional. Custom sizes downstream is a defect (in my opinion), but point heard about the impact.

geoquant (approved):

PR #729, opened 2026-08-12
chore(evals): vitest-evals Select suite + release-gated CI

mattrothenberg:
## Summary

Replaces the promptfoo-based Select evals with [`vitest-evals`](https://www.npmjs.com/package/vitest-evals). Adds an eval harness, deterministic assertions, a public-context loader, an 8-case Select suite, and a release-gated CI job that runs the evals against the "Version Packages" PR and posts a summary comment.

## What's included

- **Harness** (`evals/harness/cloudflare-ai.ts`): calls Cloudflare AI Gateway (same org secrets Bonk uses) with a direct Workers AI fallback for local dev.
- **Assertions** (`evals/assertions/select.ts`): parses generated TSX with `@babel/parser` and validates the Select API via AST. **Parse-only — generated model code is never executed.**
- **Context** (`evals/context/select.ts`): builds prompt context from public sources only (shipped registry JSON, public docs `.mdx`, public demo). No private data.
- **Suite** (`evals/select.eval.ts`) + `vitest.evals.config.ts`.
- **Scripts**: `evals`, `evals:select`, `evals:ci`, `evals:ui`.
- **CI** (`.github/workflows/release.yml`): new `evals` job, `if: github.repository_owner == 'cloudflare'`, `environment: kumo-evals`, least-privilege perms (`contents: read`, `pull-requests: write`). Uploads results artifact (30d) and comments pass/fail on the Version Packages PR.

## Safety notes

- Triggers only on `push: main` (never `pull_request`/`pull_request_target`) — forks cannot reach the AI Gateway secrets.
- Generated code is AST-parsed, not run.
- `evals/.generated/` is gitignored; result JSON verified to contain no secrets.
- Eval files typecheck clean locally. They live at repo root outside any package, so CI `test`/`typecheck` never collect them — **the eval suite will not red-X normal PR CI.**

## Open decisions for review (why this is a draft)

1. **Trigger model.** As wired, the `evals` job auto-runs (behind environment approval) on every push to `main` while a Version Packages PR is open — evals are slow/expensive, so this may be too noisy. Proposed alternative: switch to `workflow_dispatch` (manual) + `kumo-evals` approval so it runs deliberately pre-release. Feedback wanted.
2. **Environment protection (required before merge).** `kumo-evals` must have required reviewers configured, and `CF_AI_GATEWAY_*` should be **environment** secrets (not repo-wide) so the approval gate actually protects them.
3. **Token scope.** Confirm the AI Gateway token is inference-only and scoped to the eval model; consider a dedicated token separate from Bonk's for independent rotation.

## Testing

Local eval run passes 8/8. `tsc --noEmit` on all four eval files passes.

- Reviews
- [x] automated review not possible because: infra/CI change requiring human decisions on secret scoping and environment protection
- Tests
- [x] Additional testing not necessary because: eval suite is the test artifact; it runs 8/8 locally and is excluded from normal CI by design


mattrothenberg:
/bonk can you please do an initial but thorough security review of this?

mattrothenberg:
Thanks @ask-bonk — addressed the two code items in f9c553c:

**1. Error-body leak (medium)** — ✅ Fixed. Raw `body.errors` is now gated behind `KUMO_EVAL_DEBUG=1`; the default failure message only includes `response.status`, so no upstream routing/account metadata hits CI logs.

**2. Model URL interpolation (low)** — ✅ Fixed, but with a correction. The suggested `encodeURIComponent(model)` would actually **break** routing: model IDs are multi-segment paths (`@cf/moonshotai/kimi-k2.7-code`) and blanket-encoding turns the `/` separators into `%2F`. I encode **each segment** instead, preserving the slashes while neutralizing `?`/`#`/traversal chars in any override:

```ts
const encodedModelPath = model.split("/").map(encodeURIComponent).join("/");
```

**3. `kumo-evals` environment (medium / pre-merge blocker)** — Acknowledged, tracked as a pre-merge blocker in the PR description. This is a GitHub settings change (required reviewers + environment-scoped `CF_AI_GATEWAY_*` secrets), not code — will confirm before this leaves draft.

Informational items (new `vitest-evals` dep, `.env.local` DX, artifact retention) noted; no action needed for now.

PR #745, opened 2026-08-24
fix(select): anchor the Select popup below the trigger instead of overlaying it

PedroLunet:
Fixes the Select popup positioning reported on `/components/select` — the `xs` and `sm` examples opened with the popup detached from its trigger, covering adjacent rows.

> [!NOTE]
> **Reworked after review.** The first revision scaled popup option rows to match each trigger size. Per @mattrothenberg's feedback that approach is gone — **option rows are untouched**. The popup is now anchored instead, which removes the dependency between row height and placement entirely. Diff is `+203/-2` across 4 files.

## Root cause

`Select.Positioner` was rendered with **no positioning props**, so it inherited Base UI's default `alignItemWithTrigger={true}` — the native-`<select>` emulation that overlays the popup on the trigger so the selected option's text lands on the trigger's value text.

Option rows are a fixed 33px (`text-base py-1.5`). Against a 20px `xs` trigger, that overlay put the popup nowhere near its anchor. Measured on `main`:

| | `data-side` | `data-align` | popup Δx | width Δ | overhang above | overhang below | trigger covered |
|---|---|---|---|---|---|---|---|
| `xs` | `none` | `center` | −8px | +3px | 12.5px | 45.5px | yes |

`base` and `lg` were within a few px of their rows, which is why only the small sizes were visibly wrong.

## Changes

**1. The popup is anchored to a preferred side.** `Select.Positioner` now receives explicit defaults:

```tsx
side="bottom"
sideOffset={4}
align="start"
alignItemWithTrigger={false}
```

Base UI's collision handling is untouched — `side` is a *preference*, and the popup still flips automatically when that side runs out of room.

**2. Placement is configurable.** `side`, `align`, `sideOffset`, `alignOffset` and `alignItemWithTrigger` are exposed as props, documented on both `SelectPropsGeneric` and `SelectProps`. `alignItemWithTrigger` is how you opt *back into* the old native-`<select>` overlay.

**3. The popup tracks the trigger width.**

```diff
- min-w-[calc(var(--anchor-width)+3px)] py-1.5
+ max-w-(--available-width) min-w-(--anchor-width) py-1.5
```

The `+3px` existed to visually compensate for the old item-alignment shift; with anchored placement the popup is exactly trigger width.

## Verification

Measured live in Chrome DevTools. Geometry is read off the `[data-side]` popup box — note the inner `[role="listbox"]` is the scroll container and is inset by the popup's 6px `py-1.5`, so measuring that instead reports a ~10px gap.

| scenario | `data-side` | gap | width Δ | left Δ |
|---|---|---|---|---|
| `xs` / `sm` / `base` / `lg`, defaults | `bottom` | 3.9–4.25px | 0 | ≤0.5px (subpixel) |
| trigger near viewport bottom | `top` (auto-flip) | 4.06px | 0 | 0 |
| explicit `side="top"`, room available | `top` | 3.88px | 0 | 0 |
| explicit `side="top"`, no room above | `bottom` (auto-flip) | ~4px | 0 | 0 |
| explicit `side="top" sideOffset={20} align="end"` | `top` | 20px | 0 | right-aligned |
| `alignItemWithTrigger` | `none` | — | selected option centred on trigger, error **0px** |

Option row height is unchanged at 33px for every size, by design.

## Screenshots

**`xs` in the Sizes example** — the reported bug. Both frames are the same viewport, scroll position and pointer state; only the fix differs.

| Before (`main`) | After |
|---|---|
|  |  |

The popup used to swallow the `xs` trigger entirely (and the `sm` row below it); it now sits 4px underneath, left edges and width flush.

**Explicit placement** — `side="top"` renders above the trigger, still trigger-width and trigger-aligned:

**Opt-in `alignItemWithTrigger`** — the native-`<select>` overlay, kept and now documented. "Mars" is selected mid-list, so the popup extends above *and* below the trigger, with the selected row landing exactly on it:

## Docs

Two new sections on `/components/select`:

- **Placement** — `side` / `align` / `sideOffset` / `alignOffset`, four live examples.
- **Aligned to the selected option** — `alignItemWithTrigger` shown side by side with the default so the difference is visible.

## Review notes

Both of @ask-bonk's points are resolved by the rework rather than patched:

- The `py-1.25` class it flagged no longer exists — the option-scaling code is gone. (For the record it *was* valid: Tailwind v4's dynamic spacing scale emits `.py-1\.25 { padding-block: calc(var(--spacing) * 1.25) }`, which I confirmed computed to 26px in the browser.)
- `KUMO_SELECT_STYLING` is now byte-identical to `main` — this PR no longer touches it, so the "only describes `base`" nit is pre-existing code outside this diff.

---

- Reviews
- Tests
  - [ ] Tests included/updated
  - [x] Automated tests not possible - manual testing has been completed as follows: the change is popup geometry produced by Base UI's positioner at runtime, and the unit suite runs on happy-dom where `getBoundingClientRect` returns zeros, so placement cannot be asserted there. Verified instead by measuring live in Chrome DevTools (table above) across all four sizes, the automatic collision flip, three explicit placement overrides, and the `alignItemWithTrigger` overlay. Existing suite: 1267 pass / 1 pre-existing failure (`export-path-validation.test.ts`, which walks a stale local `dist/` for sourcemaps and is unrelated to `src/`). `typecheck`, `vp lint` and `vp fmt --check` all clean; docs `typecheck` clean after rebuilding the library.

### DIFF
--- a/.changeset/olive-buses-repeat.md
+++ b/.changeset/olive-buses-repeat.md
@@ -2,38 +2,28 @@
 "@cloudflare/kumo": patch
 ---
 
-fix(select): scale popup options with `size`, and add a gap when the popup falls back to anchoring
+fix(select): anchor the popup below the trigger and expose placement props
 
-Two related popup fixes. Neither changes the native-select alignment behaviour —
-`Select.Positioner` still uses Base UI's `alignItemWithTrigger`, so the selected
-option is still placed on top of the trigger whenever there is room.
+`Select.Positioner` relied on Base UI's `alignItemWithTrigger` default (`true`),
+which overlays the popup on the trigger so the selected option's text lands on
+the trigger's value text, emulating a native `<select>`. Options always render at
+`text-base`, so on `xs`/`sm` triggers the popup was much taller than its anchor,
+and Base UI's compensating shift detached it from the trigger — on `xs` it
+overhung by 12.5px above and 45.5px below and sat 8px to the left, covering
+adjacent content.
 
-**1. Options now scale with the trigger `size`.**
+The popup now prefers `bottom` placement and is offset from the trigger, with
+Base UI's collision avoidance still flipping it automatically when the preferred
+side lacks space. Placement is configurable:
 
-`Select.Option` and `Select.GroupLabel` hardcoded `text-base` / `py-1.5`, so the
-popup rendered at `base` proportions no matter what `size` the trigger used — an
-option row was a fixed 33px against a 20px `xs` trigger. Base UI shifts the
-popup to compensate for that mismatch, which visibly detached it from its
-anchor: on `xs` the popup overhung by 12.5px above and 45.5px below and was
-pushed 8px left, covering adjacent content.
+| prop | default | notes |
+| ---- | ------- | ----- |
+| `side` | `"bottom"` | preferred side; still flips automatically on collision |
+| `sideOffset` | `4` | gap between trigger and popup, matching Combobox/Autocomplete |
+| `align` | `"start"` | alignment along the trigger edge |
+| `alignOffset` | — | additional offset along the alignment axis |
+| `alignItemWithTrigger` | `false` | opt back in to the native `<select>` overlay |
 
-Option and group-label metrics now track the trigger through a size context, so
-row height matches trigger height and text indent matches trigger padding:
-
-| size | trigger | option row (before → after) | popup x-shift (before → after) |
-| ---- | ------- | --------------------------- | ------------------------------ |
-| `xs` | 20px    | 33px → 20px                 | -8px → 0                       |
-| `sm` | 26px    | 33px → 26px                 | -8px → 0                       |
-| `lg` | 40px    | 33px → 37px                 | -8px → 0                       |
-
-`base` is unchanged. A selected option now lands exactly on the trigger at every
-size instead of being offset.
-
-**2. The anchored fallback gets a 4px offset.**
-
-When there is not enough room to overlay the trigger, Base UI falls back to
-normal anchored positioning. That fallback previously rendered flush against the
-trigger; it now uses `sideOffset={4}`, matching Combobox and Autocomplete.
-
-`sideOffset` only affects the fallback — Base UI discards the anchored styles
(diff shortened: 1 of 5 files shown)


mattrothenberg (changes requested):
Let's keep the anchor offset changes, but drop the popup option scaling. Also probably makes sense for us to specify that the Select popup tries for a `bottom` placement first, but ultimately falls back to an automatic placement if necessary (or explicit user placement via prop).

PedroLunet:
Reworked per the review — thanks both. Summary of what changed since you looked:

@mattrothenberg, all four points are in:

1. **Option scaling is gone.** I reset `select.tsx` to `main` and reapplied only the positioner change, so option rows stay at 33px for every size. `SelectSizeContext` and the `py-1.25` / `KUMO_SELECT_STYLING` churn no longer exist — `KUMO_SELECT_STYLING` is byte-identical to `main` now.
2. **`sideOffset` kept** at `4`, matching Combobox/Autocomplete.
3. **Bottom-first with automatic fallback:** `side="bottom"` is a preference; Base UI's collision handling still flips it. I verified both directions — a trigger near the viewport bottom flips to `top`, and an explicit `side="top"` with no room above flips back to `bottom`.
4. **Explicit props + docs:** `side`, `align`, `sideOffset`, `alignOffset`, `alignItemWithTrigger` are all exposed and documented, with a new **Placement** section on the docs page.

The net effect is that placement no longer depends on option row height at all, which is what made the original bug possible.

I also added an **Aligned to the selected option** section documenting `alignItemWithTrigger` — the native-`<select>` overlay is still available, just opt-in now instead of the default. It's shown side by side with the anchored default, with the selection mid-list so you can see the popup extend both above and below the trigger.

@ask-bonk, for the record on the `py-1.25` finding: it was valid CSS — Tailwind v4's dynamic spacing scale emits `.py-1\.25 { padding-block: calc(var(--spacing) * 1.25) }`, which I confirmed computed to 26px in the browser. It's moot either way since that code is gone.

PR description has been rewritten with fresh measurements and screenshots. The before/after pair is captured at identical viewport, scroll position and pointer state so only the fix differs.

/bonk

mattrothenberg (approved):

PR #773, opened 2026-09-03
fix(tooltip): prevent nested text clipping

AnmolSaini16:
Related to https://gitlab.cfdata.org/cloudflare/fe/stratus/-/merge_requests/43382.

Tooltip's default trigger sets `line-height: 0`, which causes nested `Text` content that inherits line height to collapse and become clipped inside overflow-constrained consumers. Remove the trigger's line-height utility so children control their own typography, and add regression coverage for nested text.

---

- Reviews
  - [x] automated review not possible because: no automated reviewer is available in this environment
- Tests
  - [x] Tests included/updated

### DIFF
--- a/packages/kumo/src/components/tooltip/tooltip.tsx
+++ b/packages/kumo/src/components/tooltip/tooltip.tsx
@@ -171,7 +171,7 @@ export function Tooltip({
           // These prevent global button styles from polluting the trigger
           // Consumer styles passed via className will override these.
           !shouldUseRender &&
-            "m-0 inline-flex h-auto min-h-0 items-center border-none bg-transparent p-0 leading-[0] shadow-none",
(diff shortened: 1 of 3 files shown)


packages/kumo/src/components/tooltip/tooltip.tsx
```diff
@@ -171,7 +171,7 @@ export function Tooltip({
           // These prevent global button styles from polluting the trigger
           // Consumer styles passed via className will override these.
           !shouldUseRender &&
-            "m-0 inline-flex h-auto min-h-0 items-center border-none bg-transparent p-0 leading-[0] shadow-none",
```
nandanmen, line 174: I think tooltips shouldn't set a line height altogether and let the child decide the line height

AnmolSaini16: Removed the line height in the latest commit.

mattrothenberg (approved):

PR #776, opened 2026-09-04
[kumo] Add a footer to TimeseriesChart tooltips

nocdn:
### Summary

Adds a footer to the TimeseriesChart tooltip, for any extra information, or ReactNode that users might want to add.

file-cf8876b0f7dfb9106ba11022e4e55ecc

The footer works across all TimeseriesChart configurations that render a tooltip, including standard series and marker tooltips.

This PR also includes focused tests, documentation, an example, and a minor changeset.

---

- Reviews
  - [x] automated review not possible because: I do not have collaborator access required to run Bonk
- Tests
  - [x] Tests included/updated

### DIFF
--- a/packages/kumo/src/components/chart/TimeseriesChart.tsx
+++ b/packages/kumo/src/components/chart/TimeseriesChart.tsx
@@ -9,7 +9,6 @@ import {
   useMemo,
   useRef,
   useState,
-  type ReactNode,
 } from "react";
 import { Tooltip as TooltipPrimitive } from "@base-ui/react/tooltip";
 import { Chart, ChartEvents, KumoChartOption } from "./EChart";
@@ -95,11 +94,11 @@ export interface TimeseriesChartProps {
    */
   tooltipValueFormat?: (value: number) => string;
   /**
-   * Footer content rendered below the series rows in the tooltip.
+   * Footer text rendered below the series rows in the tooltip.
    * Intended for brief context such as data freshness or aggregation details.
    * Available for every `TimeseriesChart` configuration.
    */
-  tooltipFooter?: ReactNode;
+  tooltipFooter?: string;
   /**
    * Controls which series are shown in the tooltip.
    * - `"all"` — show all series at the hovered timestamp (default)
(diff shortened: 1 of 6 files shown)


packages/kumo/src/components/chart/timeseries-tooltip.tsx
```diff
@@ -29,33 +29,40 @@ export interface TooltipContentProps {
...
+          </div>
+          <SeriesTooltipRows
+            rows={state.rows}
+            hiddenCount={state.hiddenCount}
+            formatValue={formatValue}
+          />
+        </>
+      )}
+      {hasFooter && (
+        <div className="mt-1 text-xs text-kumo-subtle">{footer}</div>
```
mattrothenberg: Prefer using the `Text` component here

mattrothenberg (changes requested):

packages/kumo/src/components/chart/TimeseriesChart.tsx
```diff
@@ -93,6 +94,12 @@ export interface TimeseriesChartProps {
    * deprecated `yAxisTickLabelFormat` prop.
    */
   tooltipValueFormat?: (value: number) => string;
+  /**
+   * Footer content rendered below the series rows in the tooltip.
+   * Intended for brief context such as data freshness or aggregation details.
+   * Available for every `TimeseriesChart` configuration.
+   */
+  tooltipFooter?: ReactNode;
```
mattrothenberg: Could we limit this to just `string`? I'm always leery of exposing `ReactNode` because it's a huge vector for inconsistency.

packages/kumo/src/components/chart/timeseries-tooltip.tsx
nocdn: Sure, done in [3407a2d](https://github.com/cloudflare/kumo/commit/3407a2d86242aae0f3a80e828634b7f744ac3520)

packages/kumo/src/components/chart/TimeseriesChart.tsx
nocdn: Yes, will do, one moment

mattrothenberg (approved):

